Issue #13, opened 2019-08-12
Can't package/publish physx-sys due to generated files


repi:
Running `cargo package` on `physx-sys` current fails due to:

```sh
error: failed to verify package tarball

Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
Added: C:\git\physx-rs\target\package\physx-sys-0.1.0+4.1\PhysX\physx\include\PxConfig.h
        C:\git\physx-rs\target\package\physx-sys-0.1.0+4.1\structgen.obj

To proceed despite this, pass the `--no-verify` flag.
```

`PxConfig.h` seems to be auto-generated in case one doesn't have it, likely so all the includes can just include it. Unfortunately as we use a submodule straight from the PhysX repo we can't add one ourselves. Not sure if there are any settings to not generate or use this file. Or maybe we can fool cargo somehow as this added file is not intrusive.

`structgen.obj` I have no idea who is generating, it is some binary file and is straight in the package root which is weird and not useful for anyone.

One option can be to try and publish and get this out simply with `--no-verify` as a first step

repi:
Was able to publish still, if one has built this once then the `PxConfig.h` file will be lingering and not re-generated

PR #20, opened 2019-09-14
Fix Windows issue with generated structgen.obj file

repi:
When `build.rs` was running on Windows with MSVC the structgen.cpp file generated a structgen.obj directly in physx-sys directory during the build which fails cargo packaging & publishing.

Now I force MSVC to put object file outputs in the output path.

Also bumped the crate versions because the previous version I had published included the structgen.obj file in the crate (!) which failed to build when syncing it down because Cargo stored it (correctly) as read-only.  So I had to yank those previously published releases today.

Kinda messy situation but glad to have found a simple solution for it, this is something that was an issue from the beginning (see #13) but think it only failed packaging/publishing correctly on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 **/target
 Cargo.lock
-physx-sys/structgen.obj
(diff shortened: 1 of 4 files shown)


repi:
Yeah I think it couldn't exclude it because it was generated/touched during the build, or something like that

PR #25, opened 2019-09-23
Replace nalgebra with glam and remove ncollide

MaikKlein:


physx/src/articulation_link.rs
```diff
@@ -235,15 +235,15 @@ impl ArticulationLinkBuilder {
...
         body: &mut ArticulationReducedCoordinate,
         parent: PartHandle,
-        joint_transform: Option<glm::Mat4>,
+        joint_transform: Option<Mat4>,
     ) -> PartHandle {
-        let parent_quat = nalgebra::UnitQuaternion::from_euler_angles(
-            self.parent_rotation.x,
-            self.parent_rotation.y,
-            self.parent_rotation.z,
+        let parent_quat = Quat::from_rotation_ypr(
```
repi: Could you verify and do a test to see if the rotations are equivalent between nalgebra and glam here? glam is left-handed and may apply the rotations in a different order so something to verify

physx/src/articulation_reduced_coordinate.rs
```diff
@@ -89,10 +89,10 @@ impl ArticulationReducedCoordinate {
         let articulation =
             unsafe { PxPhysics_createArticulationReducedCoordinate_mut(physics.get_raw_mut()) };
 
-        let parent_quat = nalgebra::UnitQuaternion::from_euler_angles(
-            builder.parent_rotation.x,
-            builder.parent_rotation.y,
-            builder.parent_rotation.z,
+        let parent_quat = Quat::from_euler_angles(
```
repi: Same here with the rotations

physx/src/articulation_link.rs
MaikKlein: Yeah I'll add a few test cases. There are quite a few transformations that I have to do, I need to make sure that I am doing the correct thing.

MaikKlein:
I have fully replaced nalgebra

- [ ] Make sure everything is correct. I'll add it back in and create a few tests.
- [ ] Clean up

MaikKlein:
Rotations in glam look like they are correct. I think only perspective matrices which have a `_lh` tag are actually left handed.

physx/src/articulation_link.rs
MaikKlein: Previously this used `1 roll − 2 pitch − 3 yaw`, but the function here use yaw pitch roll.

But this is never used anyway, so I'll replace this `Vec3` which was implicitly roll pitch yaw with a `Quat`.

physx/src/math/mod.rs
```diff
@@ -26,3 +26,37 @@ impl From<PxBounds3> for Bounds {
         }
     }
 }
+
+pub struct Isometry {
+    pub translation: Mat4,
+    pub rotation: Mat4,
+}
```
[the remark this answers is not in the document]
MaikKlein: This only splits a matrix into a rotation and translation matrix. Probably a bad name for it, but better than a tuple as you don't know what is the translation/rotation.

physx/src/math/mod.rs
```diff
@@ -26,3 +26,37 @@ impl From<PxBounds3> for Bounds {
...
+impl Isometry {
+    /// Extracts the rotation and translation matrix from a matrix.
+    pub fn from_mat4(m: &Mat4) -> Self {
+        let x = m.x_axis().normalize();
+        let y = m.y_axis().normalize();
+        let z = m.z_axis().normalize();
```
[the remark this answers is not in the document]
MaikKlein: Fixed in the last commit. Panics if the axis are not of unit scale.

physx/src/math/mod.rs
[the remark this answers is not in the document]
MaikKlein: I can refactor it to use a quaternion instead of rotation matrix.

MaikKlein:
```
Nalgebra: roll pitch yaw
roll = x
pitch = y
yaw = z
                         
glam yaw pitch roll
yaw = y
pitch = x
roll = z
```

I moved away for euler angles, glam is easy to verify that it is correct but I didn't want to spent more time then necessary in nalgebra.

The latest commit added a bunch of tests to make sure that everything is correct.

physx/Cargo.toml
```diff
@@ -14,13 +14,14 @@ categories = ["api-bindings", "simulation", "game-engines"]
...
-nalgebra = "0.18.0"
-nalgebra-glm = "0.4"
-ncollide3d = "0.20.1" # TODO: Remove, tracked in https://github.com/EmbarkStudios/physx-rs/issues/17
-parking_lot = "0.9.0"
\ No newline at end of file
+glam = "0.7"
+parking_lot = "0.9.0"
+
+nalgebra-glm = "*"
+nalgebra = "*"
```
tgolsson: shouldn't these be gone now?

physx/src/math/mod.rs
```diff
@@ -10,19 +10,97 @@
...
+        //let q = Quat::from_rotation_ypr(rad(0.3), rad(0.2), rad(0.1));
+        let transform = Mat4::from_rotation_translation(q, Vec3::new(1.0, 2.0, 3.0));
+
+        let transform_nal =
+            nalgebra::Isometry3::from_parts(nalgebra_glm::vec3(1.0, 2.0, 3.0).into(), q_nal);
+
+        let transform_nal_m = transform_nal.into();
+        assert!(compare_mat(&transform, &transform_nal_m));
+    }
+}
```
tgolsson: This looks a lot better! Nice one!

physx/Cargo.toml
MaikKlein: Yes I just added some nalgebra tests. Will remove the last commit before the merge

physx/src/articulation_link.rs
```diff
@@ -235,16 +236,12 @@ impl ArticulationLinkBuilder {
         &self,
         body: &mut ArticulationReducedCoordinate,
         parent: PartHandle,
-        joint_transform: Option<glm::Mat4>,
+        joint_transform: Option<Mat4>,
```
bitshifter: Hello, author of glam here being nosy. Beware that due to 16 byte alignment putting a Mat4 in an Option wastes a lot of space.

physx/src/transform.rs
```diff
@@ -48,70 +40,45 @@ impl Into<PxQuat> for PxQuatWrap {
...
+    Vec3::new(pos.x, pos.y, pos.z)
 }
 
-/// Convert a PxVec3 to a glm::Vec3
+/// Convert a PxVec3 to a Vec3
 #[inline(always)]
-pub fn na_to_px_v3(pos: glm::Vec3) -> PxVec3 {
-    unsafe { PxVec3_new_3(pos.x, pos.y, pos.z) }
+pub fn gl_to_px_v3(pos: Vec3) -> PxVec3 {
+    unsafe { PxVec3_new_3(pos.x(), pos.y(), pos.z()) }
```
bitshifter: Might be better to do `let (x, y, z) = pos.into(); ` rather than accessing each component separately.

physx/src/transform.rs
```diff
@@ -48,70 +40,45 @@ impl Into<PxQuat> for PxQuatWrap {
...
 
-impl Into<PxTransform> for PxTransformWrap {
-    /// Unwrap helper
-    fn into(self) -> PxTransform {
-        self.0
-    }
+/// Convert a reference to a Mat4 to a PxTransform
+pub fn gl_to_px_tf_ref(trans: &Mat4) -> PxTransform {
+    let quat: PxQuatWrap = Quat::from_rotation_mat4(&trans).into();
+    let trans = trans.w_axis().truncate();
```
bitshifter: Might be better to do `let (x, y, z, _) = trans.w_axis();` rather than accessing each component separately.

physx/src/articulation_link.rs
MaikKlein: Hey, feedback is always appreciated :). And also a big thanks for writing glam, I am having a very pleasant experience so far.

Yeah 16bit alignment here seems a bit wasteful, I assume it will then become 80bytes.

MaikKlein: I think we might as well just pass in matrices directly and use `Mat4::identity` instead of `None`? @tgolsson Any thoughts?

bitshifter: Probably doesn't matter much if it's not hot code. My experience was `Mat4::inverse` originally returned `Option<Mat4>` because its idiomatic, but micro-benchmarking showed that returning as `Option` took ~80ns versus returning just a `Mat4` taking ~20ns. I tried removing the condition (e.g. always return `Some(Mat4)`) to see if that caused the slow down, but it seemed to just be the `Option`. Micro-benchmarks may not reflect real world performance, but all the same I would recommend keeping an eye on using `Option` with `Mat4` and potentially other aligned glam types in performance sensitive areas.

repi:
@MaikKlein was there anything remaining here or is this done?

MaikKlein:
All should be good, lgtm.

PR #27, opened 2019-10-02
Add github actions

MaikKlein:


repi:
Oh looks like `cmake` is not installed? That is a bit odd

MaikKlein:
Yeah weird. It lists cmake here https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions#ubuntu-1804-lts

MaikKlein:
Oh wait it just doesn't fine the cmake file which probably means build.rs doesn't do a submodule clone.

MaikKlein:
Now the build only fails for windows. I am not quite sure why though.

MaikKlein:
Seems like typeinfo got removed in the newest vs compiler :(. https://developercommunity.visualstudio.com/content/problem/734566/msvc-142328019-is-missing-include-typeinfoh.html

Jake-Shadle:
You probably want to remove the travis tests/lints as well.

PR #28, opened 2019-10-04
Update enumflags2 to 0.6

NieDzejkob:
This drops a transitive dependency on pre-1.0 `syn`, which
should improve compile times.

I wasn't able to compile the `-sys` crate locally, so I'm going to make use of your CI :)


NieDzejkob:
Ready for merging.

PR #33, opened 2019-10-31
Fix Xcode 11 support with 4.1.1

### LINKED ISSUE #23: Compile error with Xcode 11
Got a surprise update to Xcode 11 on my Mac today, which broke `physx-sys` C++ compilation:

```
PhysX/physx/source/foundation/include/unix/PsUnixIntrinsics.h:59:2: error: implicit use of sequentially-consistent atomic may incur stronger memory barriers than necessary [-Werror,-Watomic-implicit-seq-cst]
        __sync_synchronize();
        ^~~~~~~~~~~~~~~~~~
1 error generated.
```

This can be fixed by modifying the `physx-sys/PhysX/physx/source/compiler/cmake/mac/CMakeLists.txt` and adding `-Wno-atomic-implicit-seq-cst ` to the `SET(PHYSX_CXX_FLAGS` line.

Though as we bring in the PhysX SDK as a submodule we can't do that fix in this repo, and Nvidia doesn't merge PRs as they use Perforce internally so would take a while for that to get in and be able to update the submodule.

So what we would want to do is to add that compile option from our `build.rs`, tried doing something like this:

```Rust
        .define("CMAKE_CXX_FLAGS", "-Wno-atomic-implicit-seq-cst")
        .profile(build_mode)
        .build();
```

But looks like that doesn't work either, likely because the PhysX Mac CMake file just overwrites that setting: `SET(CMAKE_CXX_FLAGS ${PHYSX_CXX_FLAGS})`.

So seems we have a couple of non-ideal options:
1. Contribute a fix back to the [Nvidia repo](https://github.com/NVIDIAGameWorks/PhysX) and wait half a year for a new PhysX version to be released. Though hopefully they could merge in a fix earlier, but haven't done that before.
2. Don't submodule PhysX SDK, check in the huge SDK to this repo and modify and fix the specific file (`physx-sys/PhysX/physx/source/compiler/cmake/mac/CMakeLists.txt`).
3. Come up with another solution from our `build.rs` to make CMake add the compiler option. This would be ideal but not sure if there are any ways of doing it.

repi:
This temporary switches to use PR & fork of PhysX 4.1.1 that fixes compilation on Xcode 11:
https://github.com/NVIDIAGameWorks/PhysX/pull/185

Fix #23

### DIFF
--- a/physx-sys/build.rs
+++ b/physx-sys/build.rs
@@ -101,16 +101,8 @@ fn main() {
     println!("cargo:rustc-link-lib=static=PhysXPvdSDK_static_64");
     println!("cargo:rustc-link-lib=static=PhysXCommon_static_64");
     println!("cargo:rustc-link-lib=static=PhysXFoundation_static_64");
-
-    if "mac" == target_os {
-        // PhysX's lib names don't include the "static" segment on Mac, even though they are indeed static.
-        // See `PHYSXEXTENSIONS_LIBTYPE` in PhysX's cmake definitions for more details.
-        println!("cargo:rustc-link-lib=static=PhysXCharacterKinematic_64");
-        println!("cargo:rustc-link-lib=static=PhysXExtensions_64");
-    } else {
-        println!("cargo:rustc-link-lib=static=PhysXCharacterKinematic_static_64");
-        println!("cargo:rustc-link-lib=static=PhysXExtensions_static_64");
-    }
+    println!("cargo:rustc-link-lib=static=PhysXCharacterKinematic_static_64");
+    println!("cargo:rustc-link-lib=static=PhysXExtensions_static_64");
 
     let mut cc_builder = cc::Build::new();
     let physx_cc = cc_builder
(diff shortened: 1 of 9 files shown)


repi:
Will rebase this after #32 is merged

physx-sys/build.rs
```diff
@@ -101,16 +101,8 @@ fn main() {
     println!("cargo:rustc-link-lib=static=PhysXPvdSDK_static_64");
     println!("cargo:rustc-link-lib=static=PhysXCommon_static_64");
     println!("cargo:rustc-link-lib=static=PhysXFoundation_static_64");
-
-    if "mac" == target_os {
-        // PhysX's lib names don't include the "static" segment on Mac, even though they are indeed static.
-        // See `PHYSXEXTENSIONS_LIBTYPE` in PhysX's cmake definitions for more details.
-        println!("cargo:rustc-link-lib=static=PhysXCharacterKinematic_64");
```
h3r2tic: Aha, so they fixed the CMAKE files? Cool :)

repi: Yas!

PR #39, opened 2019-11-18
Fix cargo publish by removing PxConfig.h at the end of build.rs

h3r2tic:
PxConfig.h is generated in the source directories by PhysX's build system. This upsets cargo, so let's just delete the file after the build is done.

### DIFF
--- a/physx-sys/build.rs
+++ b/physx-sys/build.rs
@@ -235,4 +235,7 @@ fn main() {
 
     // TODO: use the cloned git revision number instead
     println!("cargo:rerun-if-changed=PhysX/physx/include/PxPhysicsVersion.h");
+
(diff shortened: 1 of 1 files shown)


physx-sys/build.rs
```diff
@@ -235,4 +235,7 @@ fn main() {
 
     // TODO: use the cloned git revision number instead
     println!("cargo:rerun-if-changed=PhysX/physx/include/PxPhysicsVersion.h");
+
+    // Remove PxConfig.h since we're only allowed to modify OUT_DIR.
+    let _ = std::fs::remove_file(physx_root_dir.join("include/PxConfig.h"));
```
repi: Could potentially do an `.expect` for it instead of ignoring the return value, to make sure the file has been removed. Though not critical

PR #45, opened 2020-02-06
Added lifetime to HeightfieldSampler so that it can capture non-static variables

Chris-Nicholls:
Without this annotation, the lifetime elision assumes that everything is captured using the 'static lifetime. This doesn't work if you want to create a function that passes in a non-static array of height values. 

For example: 

```rust
pub fn create_terrain(
    transform: Mat4,
    xsize: usize,
    ysize: usize,
    heights: &[f32],

) {

    let builder: HeightfieldBuilder = HeightfieldBuilder::default();
    let sampler: &(dyn Fn(usize, usize) -> f32) = &|x, y| heights[x*ysize+y];
    let terrain_geo = builder
        .size(xsize, ysize)
        .sampler(sampler)
        .build(&mut self.physx.cooking)
}
```

This code would fail before, with the cryptic error message that `heights` requires an explicit lifetime of 'static, but will compile with this patch.

### DIFF
--- a/physx/src/controller.rs
+++ b/physx/src/controller.rs
@@ -1,0 +1,59 @@
+#![warn(clippy::all)]
+#![warn(rust_2018_idioms)]
+
+use glam::Vec3;
+use physx_sys::{
+    PxController, PxController_getPosition, PxController_release_mut, PxController_setPosition_mut,
+    PxExtendedVec3,
+};
+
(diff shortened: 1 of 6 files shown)


Chris-Nicholls:
Closing this -  it shouldn't have been tracking from master in my fork.

PR #46, opened 2020-02-14
Do not generate _delete functions for classes with simple ctors

h3r2tic:
`_delete` should be used with classes which have `_new_alloc`; the ones using just `_new` return instances by value. Calling `_delete` on the latter would attempt to free stack memory, and result in access violations. The existence of `_delete` functions in this case is confusing and dangerous.

`_delete` functions are now only generated for the classes which use `_new_alloc`.


physx-sys/pxbind/main.cpp
```diff
@@ -1119,6 +1119,10 @@ class ClassMatchHandler : public MatchFinder::MatchCallback {
             bool hasSelf = !cmd->isStatic();
             QualType selfType = QualType(cmd->getParent()->getTypeForDecl(), 0);
 
+            const bool needsAllocatingCtor =
+                cmd->getParent()->isPolymorphic() ||
+                !cmd->getParent()->hasSimpleDestructor();
```
MaikKlein: Does that work? I am not familiar with how physx looks, but this might also need to check each field transitively.

Quickly skimming the api, there is also `bool clang::CXXRecordDecl::isPOD`.

h3r2tic: It works well enough to generate the current `_new` vs `_new_alloc` functions. IIRC, `isPOD` was way more strict.

physx-sys/build.rs
```diff
@@ -143,8 +143,14 @@ fn main() {
         .define("PX_GENERATE_STATIC_LIBRARIES", "True")
         .define("PX_GENERATE_GPU_PROJECTS", "False")
         .define("TARGET_BUILD_PLATFORM", target_os)
-        .define("CMAKE_C_COMPILER", "clang")
-        .define("CMAKE_CXX_COMPILER", "clang++")
+        .define(
+            "CMAKE_C_COMPILER",
+            env::var("CC").unwrap_or_else(|| "clang".to_owned()),
```
MaikKlein: This is a result, so you have to do `unwrap_or_else(|_| "clang".to_owned())`. ^ This does not compile

physx-sys/pxbind/main.cpp
MaikKlein: Okay lets leave it at that. The output looks goods.

PR #47, opened 2020-02-19
Added Physx format to Heightfield

Chris-Nicholls:
Replaces https://github.com/EmbarkStudios/physx-rs/pull/45 

The current heightfield code casts an array into a `*const std::ffi::c_void` that represents an array of `PxHeightFieldSample`. These happen to be 32 bits, but don't use floating-point format. It should be an array of i16s with material indices. 

This PR fixes the format. 
It assumes the heights are in the range [0..1].


tgolsson:
@Chris-Nicholls Are you finished with your changes here, or do you want to work on a bit more?

Chris-Nicholls:
@tgolsson Yup, all finished now. Thanks for being patient with the changes :-)

physx/src/heightfield.rs
```diff
@@ -8,12 +8,10 @@
...
 */
-use super::{
-    cooking::{Cooking, HEIGHT_SCALE, XZ_SCALE},
-    geometry::Geometry,
-};
+use super::{cooking::Cooking, geometry::Geometry};
 use enumflags2::BitFlags;
-use glam::Vec3;
+
+use glam::f32::Vec3;
```
tgolsson: @Chris-Nicholls Is this related to the other glam::Vec3/PxVec3 change you requested for the TriMesh Desc?

Chris-Nicholls: No, this shouldn't be here. glam::Vec3 is the same as glam::f32::Vec3

Chris-Nicholls:
```suggestion
use glam::Vec3;
```

Chris-Nicholls:
```suggestion
use glam::Vec3;
```

tgolsson: Great, thanks for the fix.

repi:
@tgolsson @h3r2tic should we get this in? or close it? 

Been sitting here for quite a while and would like us to either close it or merge it in, and do it now before we rename the `master` branch (https://github.com/EmbarkStudios/rust-ecosystem/issues/49)

tgolsson:
Sorry, yes, this fell through the cracks. I think me and @hrydgard had a discussion about this and there were some question marks with this not being quite complete, and having some odd implications. Do you remember, Henrik?

PR #48, opened 2020-02-20
Character controller

Chris-Nicholls:
Started to expose some character controller functionality.

Not all the functionality is exposed yet, just enough to create and move a capsule controller.

### DIFF
--- a/physx/src/controller.rs
+++ b/physx/src/controller.rs
@@ -4,19 +4,16 @@
 use crate::px_type::*;
 use crate::rigid_dynamic::RigidDynamic;
 
-use crate::transform::gl_to_px_v3;
 use glam::Vec3;
 use physx_macros::*;
-use physx_sys::PxControllerFilterCallback;
+
 use physx_sys::{
     phys_PxCreateControllerManager, PxCapsuleControllerDesc, PxCapsuleControllerDesc_delete,
     PxCapsuleControllerDesc_isValid, PxCapsuleControllerDesc_new_alloc, PxController,
-    PxControllerDesc, PxControllerFilters, PxControllerFilters_new, PxControllerManager,
-    PxControllerManager_createController_mut, PxController_getActor, PxController_getPosition,
-    PxController_move_mut, PxController_release_mut, PxController_setPosition_mut, PxExtendedVec3,
-    PxMaterial, PxQueryFilterCallback, PxScene,
+    PxControllerDesc, PxControllerManager, PxControllerManager_createController_mut,
+    PxController_getActor, PxController_getPosition, PxController_release_mut,
+    PxController_setPosition_mut, PxExtendedVec3, PxMaterial, PxScene,
 };
-use std::ptr::null;
 
 #[physx_type]
 impl ControllerDesc {}
@@ -66,11 +63,6 @@ impl CapsuleControllerDesc {
     }
 }
 
-pub struct Controllable {
-    controller: Controller,
-    filters: PxControllerFilters,
-}
-
 #[physx_type]
 impl Controller {
     pub fn new(controller: *mut PxController) -> Self {
@@ -95,48 +87,6 @@ impl Controller {
         unsafe {
             PxController_release_mut(self.get_raw_mut());
         }
-    }
-}
-
-impl Controllable {
-    pub fn new(controller: Controller) -> Self {
-        let filters = unsafe {
-            PxControllerFilters_new(
-                null(),
-                null::<PxQueryFilterCallback>() as *mut PxQueryFilterCallback,
-                null::<PxControllerFilterCallback>() as *mut PxControllerFilterCallback,
-            )
-        };
-        Controllable {
-            controller,
-            filters,
-        }
-    }
-
-    pub fn move_by(&mut self, disp: Vec3, min_distance: f32, elapsed_time: f32) {
-        unsafe {
-            PxController_move_mut(
-                self.controller.get_raw_mut(),
-                &gl_to_px_v3(disp),
-                min_distance,
-                elapsed_time,
-                &self.filters,
-                null(),
-            );
-        }
-    }
-}
-
-impl Deref for Controllable {
-    type Target = Controller;
-    fn deref(&self) -> &Controller {
-        &self.controller
-    }
-}
-
-impl DerefMut for Controllable {
-    fn deref_mut(&mut self) -> &mut Controller {
-        &mut self.controller
     }
 }
 
--- a/physx/src/controller.rs
+++ b/physx/src/controller.rs
@@ -21,12 +21,12 @@ impl ControllerDesc {}
 
 #[physx_type]
 impl ControllerManager {
-    pub fn new(scene: *mut PxScene) -> Self {
-        let ptr = unsafe { phys_PxCreateControllerManager(scene, false) };
+    pub fn new(scene: *mut PxScene, locking_enabled: bool) -> Self {
+        let ptr = unsafe { phys_PxCreateControllerManager(scene, locking_enabled) };
         ControllerManager { ptr }
     }
 
-    pub fn create_controller(&self, desc: &mut ControllerDesc) -> Controller {
+    pub fn create_controller(&self, desc: &ControllerDesc) -> Controller {
         let controller =
             unsafe { PxControllerManager_createController_mut(self.ptr, desc.get_raw()) };
         Controller::new(controller)
@@ -56,7 +56,11 @@ impl CapsuleControllerDesc {
                 Ok(CapsuleControllerDesc { ptr: c })
             } else {
                 PxCapsuleControllerDesc_delete(c);
-                Err(ControllerError::InvalidDescription())
+                Err(ControllerError::InvalidDescription {
+                    height,
+                    radius,
+                    step_offset,
+                })
             }
         }
     }
@@ -83,7 +87,7 @@ impl Controller {
     }
 
     pub fn get_actor(&self) -> RigidDynamic {
-        unsafe { RigidDynamic::new(PxController_getActor(self.ptr)) }
+        unsafe { RigidDynamic::from_ptr(PxController_getActor(self.ptr)) }
     }
 
     pub fn release(&mut self) {
@@ -107,8 +111,17 @@ fn from_extended(vec: PxExtendedVec3) -> Vec3 {
 
 #[derive(Error, Debug)]
 pub enum ControllerError {
-    #[error("Controller description is invalid")]
-    InvalidDescription(),
+    #[error(
+        "Controller description is invalid. height={}, radius={}, step_offset={}",
+        height,
+        radius,
+        step_offset
+    )]
+    InvalidDescription {
+        height: f32,
+        radius: f32,
+        step_offset: f32,
+    },
 
     #[error("No controller manager present")]
     NoControllerManager,
--- a/physx/src/controller.rs
+++ b/physx/src/controller.rs
@@ -14,6 +14,7 @@ use physx_sys::{
     PxControllerManager_release_mut, PxController_getActor, PxController_getPosition,
     PxController_release_mut, PxController_setPosition_mut, PxExtendedVec3, PxMaterial, PxScene,
 };
+use thiserror::Error;
 
 #[physx_type]
 impl ControllerDesc {}
@@ -27,7 +28,7 @@ impl ControllerManager {
 
     pub fn create_controller(&self, desc: &mut ControllerDesc) -> Controller {
         let controller =
-            unsafe { PxControllerManager_createController_mut(self.ptr, desc.get_raw_mut()) };
+            unsafe { PxControllerManager_createController_mut(self.ptr, desc.get_raw()) };
         Controller::new(controller)
     }
 
@@ -43,7 +44,7 @@ impl CapsuleControllerDesc {
         radius: f32,
         step_offset: f32,
         material: *mut PxMaterial,
-    ) -> Result<Self, String> {
+    ) -> Result<Self, ControllerError> {
         unsafe {
             let c = PxCapsuleControllerDesc_new_alloc();
             (*c).height = height;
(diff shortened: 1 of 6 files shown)


physx/src/controller.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn to_extended(vec: &Vec3) -> PxExtendedVec3 {
+    PxExtendedVec3 {
+        x: vec.x() as f64,
+        y: vec.y() as f64,
+        z: vec.z() as f64,
+    }
+}
+
+fn from_extended(vec: PxExtendedVec3) -> Vec3 {
```
tgolsson: Are these functions used outside the Character Controller? If so, I think they belong in either `transform.rs` or `math/mod.rs`.

physx/src/controller.rs
```diff
@@ -0,0 +1,93 @@
...
+    PxController_setPosition_mut, PxExtendedVec3, PxQueryFilterCallback, PxRigidDynamic,
+};
+use std::ptr::null;
+
+pub struct Controller {
+    controller: *mut PxController,
+    filters: PxControllerFilters,
+}
+
+impl Controller {
```
tgolsson: Can we use the `[physx_type]` macro pattern here? It's not straight forward to combine it with data ownership, unfortunately. Maybe it'd make sense to separate this into two parts, a wrapper around the PhysX-sys API (the trait) and the actual controller? It feels like the combination of these two data pieces might belong more at the application level. (I.e., in my specific character implementation for my game). 

So something like:

```rust
#[px_type]
impl Controller {
    // all API functions for PxController 
}

struct Controllable {
    controller: Controller,
    filters: ControllerFilters,
}
```

physx/src/controller.rs
```diff
@@ -0,0 +1,93 @@
...
+            );
+
+            Self {
+                controller,
+                filters,
+            }
+        }
+    }
+
+    pub fn get_raw_mut(&mut self) -> *mut PxController {
```
tgolsson: Can we implement this using the `GetRaw` trait instead?

physx/src/controller.rs
```diff
@@ -0,0 +1,93 @@
...
+        self.controller
+    }
+
+    pub fn get_raw(&self) -> *const PxController {
+        self.controller
+    }
+
+    pub fn set_position(&mut self, position: Vec3) {
+        unsafe {
+            PxController_setPosition_mut(self.get_raw_mut(), &(to_extended(&position)));
```
tgolsson: Is the extra set of parenthesis required here? (I'm guessing clippy would complain if it wasn't, but just checking since it looks odd).

physx/src/controller.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    pub fn move_by(&mut self, disp: &Vec3, min_distance: f32, elapsed_time: f32) {
```
tgolsson: I think we can take the `disp` parameter directly by value here, since the first thing we do with it is deref it.

physx/src/controller.rs
```diff
@@ -0,0 +1,93 @@
...
+                null(),
+            );
+        }
+    }
+
+    pub fn get_position(&self) -> Vec3 {
+        unsafe { from_extended(*PxController_getPosition(self.get_raw())) }
+    }
+
+    pub fn get_actor(&self) -> *mut PxRigidDynamic {
```
tgolsson: Can this return a Rust level RigidDynamic instead?

physx/src/controller.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    pub fn release(&mut self) {
+        unsafe {
+            PxController_release_mut(self.get_raw_mut());
+        }
+    }
+}
+
+impl Drop for Controller {
```
tgolsson: I'd prefer if we did not have automatic release at this level, because you can easily wreck your application:

```rust
let player_position = Controller::new(self.controller_manager.get_controller(0)).get_position(); // boom now, or maybe later
```

physx/src/scene.rs
```diff
@@ -88,6 +91,39 @@ impl Scene {
...
 
+    pub fn add_capsule_controller(
+        &mut self,
+        material: *mut PxMaterial,
+        height: f32,
+        radius: f32,
+        step_offset: f32,
+    ) -> Controller {
+        unsafe {
+            if self.controller_manager.is_none() {
```
tgolsson: I think it makes more sense to put this as part of the scene builder API, not as a runtime opt-in thing. That also lets the user decide whether to use locking.

physx/src/scene.rs
```diff
@@ -88,6 +91,39 @@ impl Scene {
...
+                self.controller_manager = Some(phys_PxCreateControllerManager(
+                    self.px_scene.write().unwrap().expect("accessing null ptr"),
+                    false,
+                ));
+            }
+
+            let c = PxCapsuleControllerDesc_new_alloc();
```
tgolsson: If you're up for it, I'd prefer if this was an actual rust struct as well, so we stick closer to the PhysX api. That also makes it easier if we end up exposing things like Materials as Rust-level types. Then we can also get rid of the panic, and let the caller decide what to do if the struct wasn't constructed correctly.

If you know, I'm also curious about the ownership here - this returns a pointer, but it's not clear to me if the controller or controller manager takes ownership of it later.

physx/src/scene.rs
tgolsson: Also, does this require any extensions to be loaded?

physx/src/controller.rs
Chris-Nicholls: No, the extended types seem to only be used for character controllers.

physx/src/controller.rs
Chris-Nicholls: 👍 Didn't see that

physx/src/controller.rs
Chris-Nicholls: Uh, no. I'll remove them.

physx/src/controller.rs
Chris-Nicholls: 👍

physx/src/controller.rs
Chris-Nicholls: Good point.

physx/src/scene.rs
Chris-Nicholls: Sure, that makes sense. Would still you leave it as optional? 
And no, this doesn't require any extensions.

physx/src/controller.rs
Chris-Nicholls: Ok, I wasn't quite sure what that macro did. I think I get it now. 

Regarding Controller/Controllable, would it make sense to use Deref and DerefMut? 
```impl Deref for Controllable {
    type Target = Controller;
    fn deref(&self) -> &Controller {
        &self.controller
    }
}

impl DerefMut for Controllable {
    fn deref_mut(&mut self) -> &mut Controller {
        &mut self.controller
    }
}
```

physx/src/scene.rs
Chris-Nicholls: Looking at it again, I don't think either take ownership - they just read the data. So it should be ok to delete it straight after using it.

Chris-Nicholls: Ok, I had a go at converting into proper rust struct. Let me know what you think!

Chris-Nicholls:
Hey @tgolsson I had a go at addressing your comments. Thanks for your time in reviewing this - I'm quite new to rust, so I appreciate you taking a look.

physx/src/controller.rs
tgolsson: No, but I'm not sure if such a `Controllable` belongs in physx-rs at all since it's not a PxSomething (right?). We only use Deref/DerefMut to emulate direct inherited function calls since there is no mechanic to do that nicely in Rust. In this case, there is nothing called a PxControllable that _is-a_ PxController as far as I know. But the `physx_type` macro can do it for the subtypes of `PxController`: 

```rust
#[physx_type(inherit = "Controller")]
impl CapsuleController {
    // PxCapsuleControllerAPI
};
```

This'll implement appropriate `GetRaw`, `Deref` and `DerefMut` for `PxType<PxCapsuleController>` and alias it as CapsuleController.

physx/src/scene.rs
tgolsson: I think having it as optional makes sense, as there's lots of use-cases where you probably don't use it. Should be fine as a `use_controller_manager(enable: bool, locking: bool)` on the builder.

physx/src/controller.rs
tgolsson: Great, then this makes sense for now!

physx/src/controller.rs
```diff
@@ -0,0 +1,103 @@
...
+#[physx_type]
+impl ControllerManager {
+    pub fn new(scene: *mut PxScene) -> Self {
+        let ptr = unsafe { phys_PxCreateControllerManager(scene, false) };
+        ControllerManager { ptr }
+    }
+
+    pub fn create_controller(&self, desc: &mut ControllerDesc) -> Controller {
+        let controller =
+            unsafe { PxControllerManager_createController_mut(self.ptr, desc.get_raw_mut()) };
```
tgolsson: The desc here should be just `get_raw()` here, it's a `*const PxControllerDesc`.

physx/src/controller.rs
```diff
@@ -0,0 +1,103 @@
...
+            let c = PxCapsuleControllerDesc_new_alloc();
+            (*c).height = height;
+            (*c).radius = radius;
+            (*c).stepOffset = step_offset;
+            (*c).material = material;
+
+            if PxCapsuleControllerDesc_isValid(c) {
+                Ok(CapsuleControllerDesc { ptr: c })
+            } else {
+                Err(format!(
```
tgolsson: Returning strings as errors is suboptimal, since it's hard to match on them on the consumer side. The most common approach is to return an enum that implements std::error::Error, like this:

```rust
pub mod controller {
    pub enum Error {
        InvalidDesc { height: f32, radius: f32, offset: f32 }
    }
    impl std::error::Error for Error {
        ...
    }

    type Result<T> = std::result::Result<T, Error>;
}

fn maybe_succeeds() -> controller::Result<Something> { ... }
```

There's a crate called `thiserror` that's commonly used to make this a bit nicer: https://docs.rs/thiserror/1.0.11/thiserror/. It should be quite straightforward to convert your String errors use this library instead, but I'll happily provide more info or help out if you want.

Also, we should free the pointer in the failure branch as well, so we don't leak it.

physx/src/scene.rs
```diff
@@ -88,6 +91,31 @@ impl Scene {
         })
     }
 
+    pub fn add_capsule_controller(
+        &mut self,
+        material: *mut PxMaterial,
+        height: f32,
+        radius: f32,
+        step_offset: f32,
+    ) -> Result<Controller, String> {
```
tgolsson: If this function takes the description instead, we don't have to return an error here. :)

physx/src/controller.rs
Chris-Nicholls: Ok, added the error enum

physx/src/scene.rs
Chris-Nicholls: Unfortunately, it still needs one for the case when no controller manager has been added.

physx/src/controller.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+fn from_extended(vec: PxExtendedVec3) -> Vec3 {
+    Vec3::new(vec.x as f32, vec.y as f32, vec.z as f32)
+}
+
+#[derive(Error, Debug)]
+pub enum ControllerError {
+    #[error("Controller description is invalid")]
+    InvalidDescription(),
```
tgolsson: This can be `InvalidDescription { height: f32,  ... }` etc, and then you can refer to those in the error just like in your original string. It makes it a bit more descriptive. :)

physx/src/controller.rs
```diff
@@ -0,0 +1,115 @@
...
+            PxController_setPosition_mut(self.get_raw_mut(), &to_extended(&position));
+        }
+    }
+
+    pub fn get_position(&self) -> Vec3 {
+        unsafe { from_extended(*PxController_getPosition(self.get_raw())) }
+    }
+
+    pub fn get_actor(&self) -> RigidDynamic {
+        unsafe { RigidDynamic::new(PxController_getActor(self.ptr)) }
```
tgolsson: use `from_ptr` here instead. They have slightly different semantics that we should probably make more apparent. `new` will allocate a Box of UserData, while `from_ptr` does not. :)

physx/src/scene.rs
```diff
@@ -503,6 +538,11 @@ impl SceneBuilder {
         self
     }
 
+    pub fn use_controller_manager(&mut self, use_controller_manager: bool) -> &mut Self {
```
tgolsson: If we could also pass in the controller manager locking flag here that'd be lovely!

physx/src/scene.rs
```diff
@@ -520,7 +560,7 @@ impl SceneBuilder {
     }
 
     /// Build a new Scene from the provided parameters
-    pub(super) fn build(&self, physics: &mut Physics) -> PxSceneDesc {
+    pub(super) fn build(&self, physics: &mut Physics) -> Scene {
```
tgolsson: This is much cleaner, thanks! :)

physx/src/scene.rs
```diff
@@ -88,6 +96,23 @@ impl Scene {
         })
     }
 
+    pub fn add_capsule_controller(
+        &mut self,
+        desc: &mut CapsuleControllerDesc,
```
tgolsson: Should be fine with a non-mutable borrow here, I think? :)

tgolsson:
@Chris-Nicholls! We're hoping to get a new release out on Thursday with this change in. Thanks! I'm hoping to review your other two PRs so they can also go out in the same version.

hrydgard:
Discovered too late that this gets rid of the only way to get a raw *PxScene, which is sometimes needed when you do stuff that the wrapper doesn't yet cover. So will have to make a followup patch release that adds get_raw to Scene.

PR #49, opened 2020-02-25
Using PxVec3 in triangle meshes, so the sizes match up

Chris-Nicholls:
* Using PxVec3 in triangle meshes, so the sizes match up
  Previously this used glam's Vec3, which does not always match the size of Physx's PxVec3. 

* Added mesh scale as an argument to `create_triange_mesh`

* Fixed validation being checked backwards

### DIFF
--- a/physx/src/cooking.rs
+++ b/physx/src/cooking.rs
@@ -100,7 +100,7 @@ impl Cooking {
         }
     }
 
-    pub fn make_geometry(&mut self, desc: ColliderDesc, mesh_scale: Vec3) -> PhysicsGeometry {
+    pub fn make_geometry(&mut self, desc: ColliderDesc) -> PhysicsGeometry {
         let geometry: Geometry = unsafe {
             match desc {
                 ColliderDesc::Sphere(radius) => Geometry::Sphere(PxSphereGeometry_new_1(radius)),
@@ -111,7 +111,11 @@ impl Cooking {
                 ColliderDesc::Cylinder(r, h) => {
                     Geometry::Capsule(PxCapsuleGeometry_new_1(r, h / 2.0))
                 }
-                ColliderDesc::TriMesh { vertices, indices } => {
+                ColliderDesc::TriMesh {
+                    vertices,
+                    indices,
+                    mesh_scale,
+                } => {
                     let mut mesh_desc = PxTriangleMeshDesc_new();
 
                     mesh_desc.points.count = vertices.len() as u32;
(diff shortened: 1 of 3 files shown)


physx/src/geometry.rs
```diff
@@ -13,8 +13,7 @@
 
 use physx_sys::*;
 
-use glam::Vec3;
-pub type Point3 = Vec3;
```
tgolsson: @Chris-Nicholls I'm hesitant about this change, since it means we're leaking a Px type in a single API. Maybe this should still use the Glam type, but the Cook handles running with SSE glam. As far as I can tell, we should be able to just change the stride in the points data passed into the cooker if we're using SSE, unless the member ordering changes, in which case we'd have to transform the input into a new vector. I don't think that's too bad since you shouldn't cook trimeshes in the hot path, but...

@repi or @h3r2tic Do you want to weigh in?

physx/src/cooking.rs
```diff
@@ -97,7 +100,7 @@ impl Cooking {
         }
     }
 
-    pub fn make_geometry(&mut self, desc: ColliderDesc) -> PhysicsGeometry {
+    pub fn make_geometry(&mut self, desc: ColliderDesc, mesh_scale: Vec3) -> PhysicsGeometry {
```
tgolsson: I'm thinking the `mesh_scale` go into the ColliderDesc for TriMesh instead, since it's only used for a single type of geometry, as none of the other ones *can* be scaled to my knowledge. Seems like a crazy source of bugs otherwise. :)

physx/src/cooking.rs
```diff
@@ -38,10 +38,14 @@ impl Cooking {
...
-    pub fn create_triangle_mesh(&self, mesh_desc: &PxTriangleMeshDesc) -> Result<Geometry, ()> {
+    pub fn create_triangle_mesh(
+        &self,
+        mesh_desc: &PxTriangleMeshDesc,
+        mesh_scale: Vec3,
+    ) -> Result<Geometry, ()> {
         let mut cooking_result = PxTriangleMeshCookingResult::eSUCCESS;
         unsafe {
-            if self.validate_triangle_mesh(mesh_desc) {
+            if !self.validate_triangle_mesh(mesh_desc) {
```
tgolsson: Ouch, that is... I'm wondering if this has ever worked. Nice catch!

physx/src/cooking.rs
Chris-Nicholls: That's a much better idea

physx/src/geometry.rs
hrydgard: Maybe just fall back to [f32; 3] as the trimesh vertex type?   They can be trivially converted to and from glam Vec3 using into() anyway, so aren't too inconvenient

PR #50, opened 2020-03-03
Make calling PxDefaultSimulationFilterShader from the trampoline optional.

hrydgard:


physx-sys/src/lib.rs
```diff
@@ -155,10 +155,12 @@ extern "C" {
     /// Deallocates the PxSimulationEventCallback that has previously been created
     pub fn destroy_contact_callback(callback: *mut PxSimulationEventCallback);
 
-    /// Sets a trampoline which will invoke *both* the default filter shader for the pair flags
-    /// *as well as* the custom filter shader which can add pair flags and set the collision handler flag
+    /// Sets a trampoline which will, if call_default_filter_shader_first is true, invoke *both* the default
+    /// filter shader for the pair flags *as well as* the custom filter shader which can add pair flags and
+    /// set the collision handler flag. Or if it's false, just the latter.
```
tgolsson: This is now a three-line long sentence. Can we simplify this to: 

> Override the default filter shader in the scene with a custom function. If `call_default_filter_shader_first` is set to true, this will first call the built-in PhysX filter before your callback.

hrydgard: Done!

PR #52, opened 2020-03-03
Allow sneaking in a userdata pointer into the filter shader callback.

hrydgard:


physx/src/scene.rs
```diff
@@ -517,6 +519,15 @@ impl SceneBuilder {
         self
     }
 
+    /// Sets a userdata pointer that will be carried into the filter shader callback function,
+    /// through the constantBlock parameter.
+    ///
+    /// Default: null
+    pub fn set_filter_shader_userdata(&mut self, user_data: *mut std::ffi::c_void) -> &mut Self {
+        self.filter_shader_user_data = user_data;
```
tgolsson: This feels like a footgun. This *should* probably be both Sync and Pinned for safety reasons, but I'm not sure how we can do that.

hrydgard: Total footgun. When I needed something similar before I put something in a Box and returned a pointer to that, it worked fine though :P

hrydgard:
I don't actually think I need this right now (except for an optimization I don't care about) so can hold off on this if we want to think more about the footgun danger this introduces.

tgolsson:
> I don't actually think I need this right now (except for an optimization I don't care about) so can hold off on this if we want to think more about the footgun danger this introduces.

I think we have to think about this kind of thing a lot. The problem is that any semantic we can introduce where we use Rusts guarantees also requires us to manage lifetimes; which makes the API real messy. The userdata here should be `T: Pin<T> + Sync + 'a` and the whole Scene is then `Scene<'a>` as it cannot outlive the things it has callbacks for. This is obviously *sane*, but it gets very close to syntax hell. 

I'm not sure how the lifetime inference deals with it but iirc it's OK to do something like

```rust
struct PhysicsState {
    scene: Scene,
}
```

and then it's accepted to pass `&mut Self` into `self.scene` and the lifetimes are OK because by definition the scene cannot outlive its container.

hrydgard:
Closing this to revisit in the future once we thought more about userdata and callbacks.

PR #54, opened 2020-03-04
Release physx 0.5.0 and physx-sys 0.3.0

hrydgard:
Should we squash?

Had to make the last commit myself.


hrydgard:
Oh.. Maybe we shouldn't let it do those alpha version thingies. CI didn't like that but I don't want to specify exactly either? Hm.

tgolsson:
> Should we squash?

I think this is OK; but maybe kick out the last two commits? Not sure. Seems like they break CI...

hrydgard:
Thanks! Retagged, too.

PR #59, opened 2020-03-18
Enable Cross Compilation

Jake-Shadle:
The biggest change is that cmake is now an optional method of building that can be enabled with the `use-cmake` feature flag. However, by default, [`cc`](https://github.com/alexcrichton/cc-rs) will be used (with the parallel feature) for compilation. This was the easiest way to get physx to compile for windows uses clang-cl on a non-Windows host platform.

In addition, the structgen executable, which is also built and executed by the build script, is also behind a feature flag, `structgen`, and will not be compiled or executed by default. Instead, the outputs for linux, macos, and windows have been checked in to the repository which means they can be used when running on a platform other than the target.

Changes were made to the CI to run tests and the ball example in debug and release with all, none, only structgen, and only cmake feature flags on to make sure that the various scenarios all "work".

Note that this supplants #41 as it takes into account the CXX compiler in the same way that cc-rs does.
Resolves: #41 

TODO: Put cooking and the visual debugger behind feature flags as they're not required for the core physx functionality


Jake-Shadle:
Oh right, and I will update the README/docs once this gets some more eyes on it.

physx-sys/build.rs
```diff
@@ -1,166 +1,83 @@
...
     };
 
-    let target_os = get_physx_target_os_name();
-    let target_os = target_os.as_str();
-    let mut physx_cfg = Config::new("PhysX/physx/source/compiler/cmake");
-
-    if target_os == "linux" {
-        physx_cfg.define("CMAKE_LIBRARY_ARCHITECTURE", "x86_64-unknown-linux-gnu");
-    }
+    let use_cmake = env::var("CARGO_FEATURE_USE_CMAKE").is_ok();
```
h3r2tic: This is potentially a bit confusing -- other cargo/rust flags use the `=0` and `=1` convention to disable and enable a feature. Here you would need to actually unset the var.

physx-sys/build.rs
```diff
@@ -173,69 +90,94 @@ fn main() {
...
+        .file("src/physx_api.cpp")
         .compile("physx_api");
 
     println!("cargo:rerun-if-changed=src/structgen/structgen.cpp");
     println!("cargo:rerun-if-changed=src/structgen/structgen.hpp");
     println!("cargo:rerun-if-changed=src/lib.rs");
     println!("cargo:rerun-if-changed=src/physx_generated.hpp");
     println!("cargo:rerun-if-changed=src/physx_generated.rs");
-    println!("cargo:rerun-if-changed=src/physx.cpp");
+    println!("cargo:rerun-if-changed=src/physx_apicpp");
```
h3r2tic:
```suggestion
    println!("cargo:rerun-if-changed=src/physx_api.cpp");
```

physx-sys/build.rs
Jake-Shadle: This is just due to how features appear in the environment.

> `CARGO_FEATURE_<name>` — For each activated feature of the package being built, this environment variable will be present where <name> is the name of the feature uppercased and having - translated to _.

h3r2tic: Ah, I see. So one would never set this by hand. Cool, sorry for the confusion then.

h3r2tic:
Being able to build this without CMAKE, and to cross-compile is fantastic, thanks :D

The code looks good to my eye, but one thing is unclear: how do we update to a newer version of PhysX? What's the path for upgrading `structgen` and the lists of source files?

physx-sys/build.rs
Jake-Shadle: No worries, glad you're keeping me honest!

Jake-Shadle:
So I came into this without knowing anything about the physx code base or how it evolves, so I was planning on doing the next upgrade myself so I could writeup complete documentation for it. I wasn't overly concerned with it due their extremely long release cadence though, so there are a few manual steps (like downloading the generated files from CI for all platforms manually to place them in the repo).

h3r2tic:
Works for me! Just don't get hit by a train, ktnx! :D

Jake-Shadle:
So far so good!

PR #61, opened 2020-05-07
Release physx 0.6.0 and physx-sys 0.4.0

repi:
Biggest thing this includes is #59 and #60.

Made this not a patch version release, as although those changes are compatible they are fairly major in the build system so better for users to choose to upgrade to them manually with a version change.


repi:
Ah nice, forgot we had the nice CHANGELOGs now!

Jake-Shadle:
I did too, at least when I made the change in the first place. :sweat_smile:

PR #66, opened 2020-05-18
Android Support

VZout:
Initial android support.

Added casts to `base.rs`, `rigid_static.rs` and `visual_debugger.rs`. The internal pointer type seems to differ for android. However this should not be a problem if I cast using `as _`.

*Ignore the commit history*


Jake-Shadle:
I realize this is just a draft, but I was just wondering if Android has some kind of standard/de facto standard way to communicate the SDK path to things that need it? Obviously the path is hardcoded right now for iteration, but that will have to change at some point. It would be a shame if users were forced to set and environment variable before running the build if they are targeting android.

hrydgard:
$ANDROID_HOME should point to the SDK if it's installed correctly.

VZout:
> $ANDROID_HOME should point to the SDK if it's installed correctly.

Yep!

VZout:
> $ANDROID_HOME should point to the SDK if it's installed correctly.

In this case I want `NDK_HOME` I believe. ANDROID_HOME should point to the Sdk.

Jasper-Bekkers:
Minor side-note: this will also require `qemu-aarch64` to be installed on the CI systems @Jake-Shadle

physx-sys/pxbind/run_android.sh
```diff
@@ -0,0 +1,12 @@
+cd ../src
+export PHYSX_ROOT_DIR=../PhysX/physx
+../pxbind/build/pxbind ../PhysX/physx/include/PxPhysicsAPI.h -- \
+    --target=aarch64-linux-android \
+    -DNDEBUG=1 -x c++ \
+    -DDISABLE_CUDA_PHYSX \
+    -I../pxbind \
+    -I../PhysX/physx/include \
+    -I../PhysX/physx/../pxshared/include/ \
+    -I/home/viktor/android-ndk-r21b/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1 \
```
Jasper-Bekkers, line 10: Probably should be $NDK_HOME instead.

VZout: Ah yeah. That line needs to removed completely

repi:
Sweet!

Jake-Shadle:
I think it is probably time to split the Linux image at this point to have a few of them for different things.

Jake-Shadle:
Wait wrong repo. 😅

repi:
Can be a separate issue, but can we verify Android in CI that it builds for it and ideally runs our example or tests also?

hrydgard:
Fantastic!

Jasper-Bekkers:
> Can be a separate issue, but can we verify Android in CI that it builds for it and ideally runs our example or tests also?

Filed issue here: #68

PR #72, opened 2020-06-29
Upgrade glam to 0.9

hrydgard:
Replaces #71 .

Before we merge this, I'll patch my local ark to use this and test that it works.


hrydgard:
It works.

hrydgard:
Done, I think we'll need to bump the "major" version btw...

repi:
Yeah has to be a major version as the glam types are part of the public API

hrydgard:
@repi This needs another approval to merge now

PR #73, opened 2020-06-30
Add the ability to register a raycast filter callback from Rust

hrydgard:
With this, you can register a callback to use to decide what objects can be hit by raycasts. In the callback, you get the actor, the custom filterdata for the mesh, and a userdata pointer. To reject a single object for example, pass in the PxActor through userdata and compare.

### DIFF
--- a/physx-sys/src/physx_api.cpp
+++ b/physx-sys/src/physx_api.cpp
@@ -118,6 +118,32 @@ class RaycastFilterCallback : public PxQueryFilterCallback
     }
 };
 
+typedef int (*RaycastHitCallback)(const PxRigidActor *actor, PxFilterData const *filterData, const void *userData);
+
+class RaycastFilterTrampoline : public PxQueryFilterCallback
+{
+  public:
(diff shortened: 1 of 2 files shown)


physx-sys/src/physx_api.cpp
```diff
@@ -118,6 +118,32 @@ class RaycastFilterCallback : public PxQueryFilterCallback
     }
 };
 
+typedef int (*RaycastHitCallback)(const PxRigidActor *actor, PxFilterData const *filterData, const void *userData);
```
h3r2tic, line 121: I think it might be a good idea to have all the parameters that `PxQueryFilterCallback::preFilter` receives, so that if we find out in a while that we need to use the `PxShape` or `PxHitFlags`, we don't end up needing to create another callback type.

hrydgard: Yeah, fair enough.

hrydgard:
All the parameters are in now. Made the flags one an u32 for FFI purposes...

PR #74, opened 2020-06-30
Fix Android NDK toolchain path selection on Mac & Windows

repi:
It was hardcoded to `linux-x86_6`, so here detect which target the host is to pick the right toolchain for Windows and Mac.

Also manually verify that the sysroot path for the target actually exists as clang will just ignore it if it doesn't exist, and then you'll get "fun" compile errors instead


repi:
Oh in later v21 of the NDK the toolchain path for windows is called `windows-x86_64` instead of just `windows` (which I thought was odd). Will test with the latest version locally first and then change this PR to reflect that

repi:
Fixed

PR #79, opened 2020-07-23
Add support for aarch64-unknown-linux-gnu

repi:


Nehliin:
fixed in https://github.com/EmbarkStudios/physx-rs/pull/154

PR #80, opened 2020-08-12
New interface for registering simulation event callbacks

hrydgard:
Now supporting registering all six types, not just collision. Needed for trigger shapes.

Preserves, and deprecates, the old interface.

### DIFF
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -268,10 +268,10 @@ extern "C" {
     pub fn destroy_contact_callback(callback: *mut PxSimulationEventCallback);
 
     /// New interface to handle simulation events, replacing create_contact_callback.
-    pub fn create_simulation_event_handler(
+    pub fn create_simulation_event_callbacks(
         callbacks: *const SimulationEventCallbackInfo,
     ) -> *mut PxSimulationEventCallback;
-    pub fn destroy_simulation_event_handler(callback: *mut PxSimulationEventCallback);
+    pub fn destroy_simulation_event_callbacks(callback: *mut PxSimulationEventCallback);
 
     /// Override the default filter shader in the scene with a custom function.
     /// If call_default_filter_shader_first is set to true, this will first call the
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -171,6 +171,48 @@ pub const fn version(major: u32, minor: u32, patch: u32) -> u32 {
 pub type CollisionCallback =
     unsafe extern "C" fn(*mut c_void, *const PxContactPairHeader, *const PxContactPair, count: u32);
 
+pub type TriggerCallback = unsafe extern "C" fn(*mut c_void, *const PxTriggerPair, count: u32);
+
+pub type ConstraintBreakCallback =
+    unsafe extern "C" fn(*mut c_void, *const PxConstraintInfo, count: u32);
+
+pub type WakeSleepCallback =
+    unsafe extern "C" fn(*mut c_void, *const *const PxActor, count: u32, bool);
+
+pub type AdvanceCallback =
+    unsafe extern "C" fn(*mut c_void, *const *const PxRigidBody, *const PxTransform, count: u32);
+
+#[repr(C)]
+pub struct SimulationEventCallbackInfo {
+    pub collision_callback: Option<CollisionCallback>, // Yes, option<funcptr> becomes a nullable C pointer
+    pub collision_user_data: *mut c_void,
+    pub trigger_callback: Option<TriggerCallback>,
+    pub trigger_user_data: *mut c_void,
+    pub constraint_break_callback: Option<ConstraintBreakCallback>,
+    pub constraint_break_user_data: *mut c_void,
+    pub wake_sleep_callback: Option<WakeSleepCallback>,
+    pub wake_sleep_user_data: *mut c_void,
+    pub advance_callback: Option<AdvanceCallback>,
+    pub advance_user_data: *mut c_void,
+}
+
+impl Default for SimulationEventCallbackInfo {
+    fn default() -> Self {
+        Self {
+            collision_callback: None,
+            collision_user_data: std::ptr::null_mut(),
+            trigger_callback: None,
+            trigger_user_data: std::ptr::null_mut(),
+            constraint_break_callback: None,
+            constraint_break_user_data: std::ptr::null_mut(),
+            wake_sleep_callback: None,
+            wake_sleep_user_data: std::ptr::null_mut(),
+            advance_callback: None,
+            advance_user_data: std::ptr::null_mut(),
+        }
+    }
+}
+
 /// return 0 = PxQueryHitType::eNONE
 /// return 1 = PxQueryHitType::eTOUCH
 /// return 2 = PxQueryHitType::eBLOCK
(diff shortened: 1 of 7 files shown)


physx-sys/src/physx_api.cpp
```diff
@@ -52,45 +52,73 @@ PxFilterFlags FilterShaderTrampoline(PxFilterObjectAttributes attributes0,
     return PxFilterFlags{shaderfilter(&info)};
 }
 
-class CollisionFilterTrampoline : public PxSimulationEventCallback
+typedef void (*CollisionCallback)(void *, PxContactPairHeader const *, PxContactPair const *, PxU32);
+typedef void (*TriggerCallback)(void *, PxTriggerPair const *, PxU32);
+typedef void (*ConstraintBreakCallback)(void *, PxConstraintInfo const *, PxU32);
+typedef void (*WakeSleepCallback)(void *, PxActor **const, PxU32, bool);
+typedef void (*AdvanceCallback)(void *, const PxRigidBody *const *, const PxTransform *const, PxU32);
```
tgolsson, line 59: I'd argue these are much more readable as using declarations!

```c++
using CollisionCallback = void(*)(void*, PxContactPairHeader const *, PxContactPair const *, PxU32);
...
```

physx-sys/src/lib.rs
```diff
@@ -171,6 +171,48 @@ pub const fn version(major: u32, minor: u32, patch: u32) -> u32 {
...
+pub type TriggerCallback = unsafe extern "C" fn(*mut c_void, *const PxTriggerPair, count: u32);
+
+pub type ConstraintBreakCallback =
+    unsafe extern "C" fn(*mut c_void, *const PxConstraintInfo, count: u32);
+
+pub type WakeSleepCallback =
+    unsafe extern "C" fn(*mut c_void, *const *const PxActor, count: u32, bool);
+
+pub type AdvanceCallback =
+    unsafe extern "C" fn(*mut c_void, *const *const PxRigidBody, *const PxTransform, count: u32);
```
tgolsson, line 183: Why is only the count named above? Is it required to be? It looks odd.

physx-sys/src/lib.rs
```diff
@@ -171,6 +171,48 @@ pub const fn version(major: u32, minor: u32, patch: u32) -> u32 {
...
+    pub collision_user_data: *mut c_void,
+    pub trigger_callback: Option<TriggerCallback>,
+    pub trigger_user_data: *mut c_void,
+    pub constraint_break_callback: Option<ConstraintBreakCallback>,
+    pub constraint_break_user_data: *mut c_void,
+    pub wake_sleep_callback: Option<WakeSleepCallback>,
+    pub wake_sleep_user_data: *mut c_void,
+    pub advance_callback: Option<AdvanceCallback>,
+    pub advance_user_data: *mut c_void,
+}
```
tgolsson, line 197: I'd maybe add some description here of these fields and maybe just space them out a bit! :-)

tgolsson: Also, what's the semantic difference between the context being a ptr and a nullable function ptr? I.e., could we represent the latter as `Option<NonNull<Any>>`? Or if the context always must exist, just `NonNull<Any>`? Or a box? :-) Trying to think if/how we can be more clear about semantics.

physx-sys/src/physx_api.cpp
```diff
@@ -52,45 +52,73 @@ PxFilterFlags FilterShaderTrampoline(PxFilterObjectAttributes attributes0,
...
+
+struct SimulationEventCallbackInfo {
+    CollisionCallback collisionCallback;
+    void *collisionUserData;
```
tgolsson, line 63: These should probably default to nullptr as well.

physx-sys/src/physx_api.cpp
```diff
@@ -52,45 +52,73 @@ PxFilterFlags FilterShaderTrampoline(PxFilterObjectAttributes attributes0,
...
 {
   public:
-    CollisionFilterTrampoline(CollisionCallback callback, void *userData) : mUserData(userData), mCallback(callback)
-    {}
-
-    void onContact(const PxContactPairHeader &pairHeader, const PxContactPair *pairs, PxU32 nbPairs) override
-    {
-        mCallback(mUserData, &pairHeader, pairs, nbPairs);
+    SimulationEventTrampoline(const SimulationEventCallbackInfo *callbacks) {
+        mCallbacks = *callbacks;
```
tgolsson, line 78: Do this in the initializer list instead, not in the body.

tgolsson: To be fair it really irks me overall that this is just copying the whole struct - I get why; but this is essentially/potentially a transfer of ownership to this callback... I'm not sure if we can use lifetimes here but really the SimulationEventCallbackInfo has to have a lifetime at least as long as the trampoline itself... Which are both less than the full scene. Hmm. I think we've had this discussion before; and I still don't know how to represent it in Rust terms.

physx-sys/src/physx_api.cpp
hrydgard: I guess it depends on what you're used to, but fine by me :)

physx-sys/src/lib.rs
hrydgard: No, it's not. Just cargo culting myself. Removed it.

physx-sys/src/lib.rs
hrydgard: Clarified it with a new comment: 

// Function pointers in Rust are normally not nullable (which is why they don't require unsafe to call)
// but we need them to be, so we simply wrap them in Option<>. An Option<funcptr> is luckily represented
// by the compiler as a simple pointer with null representing None, so this is compatible with the C struct.

physx-sys/src/physx_api.cpp
hrydgard: Oops.

physx-sys/src/physx_api.cpp
hrydgard: It's a small struct and it's easier to just copy it. This is C++ so no lifetime annotations...

tgolsson: I know it's C++, my point was more that I'd want this to be safer and be more transparent w.r.t. the expected lifetimes on the Rust side - making the API safer. :-) Maybe something funky like

```rust
enum CallbackPayload {
    OnWake(...),
    OnContact(...),
}

struct SimulationCallback<'a, WakeUserData, ContactUserData, ...> {
     callback_trampoline: *mut PxSimulationCallback,
     wake_data: SomeRefType<'a, WakeUserData>,
     ...
}

impl SimulationCallback<...> {
    fn on_callback(self: *mut Self, data: *const CallbackPayload) { ... }
}
```

And then ~ish the callback trampoline is tied to some lifetime in Rust land... Just thinking out loud.

hrydgard: Figuring out how to do this stuff in a safe or safer manner is not something I have time or energy for right now though, unfortunately. There is heavy demand for trigger volumes and a bunch more features and this works for now.

After all this is in place, I want to go back and move all the unsafe stuff we've been doing in Ark into safe nice wrappers here, but it hasn't been possible to do it continuously since the speed of change and discovery of more stuff we need from PhysX has just been too high. We need actual usage to know what the safe interfaces should look like anyway.

tgolsson: Yeah, I understand that! Approved now.

PR #81, opened 2020-08-12
Add missing lint to CHANGELOG.

hrydgard:
Dummy PR to check CI on main, really.


hrydgard:
@repi  as expected, merging that broke the build :)  This was only to test CI.

PR #83, opened 2020-08-13
Release physx 0.7.1 and physx-sys 0.4.9

repi:
@hrydgard looks like you deprecated `set_simulation_event_callback` also in #80, can you add that to the CHANGELOG also. I guess technically we should do a minor release rather than a patch release if it has deprecations, but think this is fine as we are pre 1.0

### DIFF
--- a/physx-sys/CHANGELOG.md
+++ b/physx-sys/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## [Unreleased]
 
+## [0.4.9] - 2020-08-13
+
 ### Added
 
 - [PR#80](https://github.com/EmbarkStudios/physx-rs/pull/80) New way to register simulation event callbacks, supporting all callback types this time.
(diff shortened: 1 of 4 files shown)


Jake-Shadle:
No I think deprecations in a patch are fine, but it would of course need a minor bump when the function was actually removed.

repi:
> No I think deprecations in a patch are fine, but it would of course need a minor bump when the function was actually removed.

It is not ideal to do it in patch releases as it is not nice to do a `cargo update` (or just building without a lockfile) and all the sudden you are getting deprecation warnings. 

Also looked up what semver.org says, and it specifically wants it to be in minor version changes:

https://semver.org/#spec-item-7
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

That said though, as we are pre 1.0 and with Rust's 0.x versioning I think it is fine

Jake-Shadle:
Ahh ok, makes sense especially if you have warnings as errors.

hrydgard:
I added a commit to this branch with the deprecation note in the CHANGELOG.

PR #85, opened 2020-09-05
Provide an API endpoint to allow users to remove actors from the scene

### LINKED ISSUE #84: Removing an actor from a scene fails and prints a PhysX warning instead
**Describe the bug**
Calling physics::remove_body on a BodyHandle returned by scene::add_actor or scene::add_dynamic will fail and print out the following error instead:
`C:\physx-rs\physx-sys\PhysX/physx\source/physx/src\NpScene.cpp (79) : invalid operation : PxScene::removeArticulation(): Articulation not assigned to scene or assigned to another scene. Call will be ignored!`
This is because physics::remove_body calls scene::remove_articulation internally, regardless of whether or not the BodyHandle argument points to an articulation.

**To Reproduce**
1. Add an actor to the scene using scene::add_actor or scene::add_dynamic and save the returned BodyHandle.
2. Given this handle, attempt to remove the actor from the scene by calling physics::remove_body.
3. Observe that the actor is not removed from the scene and an "invalid operation" is reported.

**Expected behavior**
The API should expose a mechanism to delete actors and not just articulations. It should be difficult or impossible to call removeArticulation on an actor or call removeActor on an articulation.

**Screenshots**
n/a

**Device:**
 - OS: n/a
 - Browser: n/a
 - Version: origin/main

**Additional context**
First, I want to say thank you for all the effort put into this crate! You all are awesome, and this crate clearly has a lot of effort put into it. The associated talk is great too. :)

I've been toying around with physx-rs for a basic first-person shooter lately and ran into this issue. I actually have a pull request associated with this issue that I'll add here shortly. It's not a perfect solution, but it's a step in the right direction, and I'm hoping that we can refine it where needed. My project focuses on static and dynamic actors rather than articulations, so I could really use some tests that verify that my work does not break articulation API, at least until I start toying with those too.

NickCaplinger:
Fixes #84 

Hi all. The relevant information for this pull request can be found in the issue above and the commit message for 
6a96fe3. I'm very much open to discussion on this. Thanks for the crate again!

Cheers!
Nick

### DIFF
--- a/physx/src/articulation_base.rs
+++ b/physx/src/articulation_base.rs
@@ -179,35 +179,34 @@ impl ArticulationBase {
         transform: Option<Mat4>,
         joint_transform: Option<Mat4>,
     ) -> *mut PxArticulationLink {
+        // FIXME Assumes that self is the correct articulation for parent, which may not be true
 
-		// FIXME Assumes that self is the correct articulation for parent, which may not be true
+        if let ActorType::ArticulationLink(_articulation_handle) = parent.1 {
+            let (parent, trans) = if parent.0 == 0 {
+                (null_mut(), Mat4::identity())
+            } else {
+                let parent_link = parent.0 as *mut PxArticulationLink;
+                (parent_link, unsafe {
+                    px_to_gl_tf(PxRigidActor_getGlobalPose(
+                        parent_link as *const PxRigidActor,
+                    ))
+                })
+            };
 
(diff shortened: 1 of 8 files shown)


NickCaplinger:
It looks like the Windows tests are failing due to linker errors, but I don't think it's related to my change.

NickCaplinger:
Fixed by a recent redesign of the crate. See more details here: https://github.com/EmbarkStudios/physx-rs/issues/84

PR #87, opened 2020-10-12
Allow setting the solver type when creating a scene

hrydgard:
Also makes the broadphasetype setting work, it never did anything.

Additionally, disables a meta-warning that has appeared in newer Rust compilers:

```
warning: warn(rust_2018_idioms) is ignored unless specified at crate level
 --> C:\embark\physx-rs\physx\src\px_type.rs:6:9
  |
6 | #![warn(rust_2018_idioms)]
  |         ^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_attributes)]` on by default
```

### DIFF
--- a/physx/src/lib.rs
+++ b/physx/src/lib.rs
@@ -4,6 +4,7 @@
 
 #![warn(clippy::all)]
 #![warn(rust_2018_idioms)]
+#![allow(unused_attributes)]
 
 //! # 🎳 physx
 //!
--- a/physx/src/lib.rs
+++ b/physx/src/lib.rs
@@ -4,7 +4,6 @@
 
 #![warn(clippy::all)]
 #![warn(rust_2018_idioms)]
-#![allow(unused_attributes)]
 
 //! # 🎳 physx
 //!
--- a/physx/src/scene.rs
+++ b/physx/src/scene.rs
@@ -3,7 +3,6 @@
 // Created: 10 April 2019
 
 #![warn(clippy::all)]
-#![warn(rust_2018_idioms)]
 #![allow(clippy::missing_safety_doc)]
 #![allow(deprecated)]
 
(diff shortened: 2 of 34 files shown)


physx/src/scene.rs
```diff
@@ -628,6 +646,14 @@ impl SceneBuilder {
         self
     }
 
+    /// Set solver type
+    ///
+    /// Default: eTGS
```
tgolsson, line 651: Wrong default value!

physx/src/scene.rs
```diff
@@ -645,6 +671,9 @@ impl SceneBuilder {
 
             scene_desc.cpuDispatcher = dispatcher;
             scene_desc.gravity = gl_to_px_v3(self.gravity);
+            scene_desc.solverType = self.solver_type.into();
+            scene_desc.broadPhaseType = self.broad_phase_type.into();
```
tgolsson, line 675: Oh, nice catch.

physx/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 
 #![warn(clippy::all)]
 #![warn(rust_2018_idioms)]
+#![allow(unused_attributes)]
```
tgolsson, line 7: I'd remove the preceding line instead - it does nothing where it is.

tgolsson (changes requested):

physx/src/lib.rs
hrydgard: The reason I added this was that `#![warn(rust_2018_idioms)]` is present in 32 files. I'll delete it in all of them instead.

tgolsson (approved):
Approved when build is green!

hrydgard:
Seeing a bizarre error on the Windows tests on CI: 

```
PhysXCooking_static_64.lib(PhysXCooking.res) : fatal error LNK1241: resource file PhysX_static_64.lib(PhysX.res) already specified
```

PR #90, opened 2020-10-14
Expose a bunch of additional parameters on SceneBuilder

hrydgard:
Pretty straightforward.

### DIFF
--- a/physx/src/scene.rs
+++ b/physx/src/scene.rs
@@ -548,6 +548,11 @@ pub struct SceneBuilder {
     pub(crate) use_ccd: bool,
     pub(crate) enable_ccd_resweep: bool,
     pub(crate) solver_type: SolverType,
+    pub(crate) enhanced_determinism: bool,
+    pub(crate) friction_every_iteration: bool,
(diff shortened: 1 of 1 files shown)


physx/src/scene.rs
```diff
@@ -563,6 +568,11 @@ impl Default for SceneBuilder {
             use_ccd: false,
             enable_ccd_resweep: false,
             solver_type: SolverType::PGS,
+            enhanced_determinism: false,
+            friction_every_iteration: false,
+            stabilization: false,
+            adaptive_force: false,
+            ccd_max_passes: 1,
```
tgolsson, line 575: Are these the default settings for PhysX?

emilk (approved):
🥳

tgolsson (approved):
LGTM with the caveat below.

physx/src/scene.rs
hrydgard: Yes.

tgolsson: Perfect!

PR #94, opened 2020-10-20
Allow overriding the PxAllocatorCallback with a pair of custom callback functions

hrydgard:
This lets the user handle or track allocations as they please.

### DIFF
--- a/physx-sys/src/physx_api.cpp
+++ b/physx-sys/src/physx_api.cpp
@@ -177,11 +177,39 @@ class RaycastFilterTrampoline : public PxQueryFilterCallback
     }
 };
 
+typedef void * (*AllocCallback)(uint64_t size, const char *typeName, const char *filename, int line, void *userdata);
+typedef void (*DeallocCallback)(void *ptr, void *userdata);
+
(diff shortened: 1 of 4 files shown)


physx-sys/src/physx_api.cpp
```diff
@@ -212,6 +240,14 @@ extern "C"
         return new RaycastFilterTrampoline(callback, userData);
     }
 
+    PxAllocatorCallback *create_alloc_callback(
+        AllocCallback alloc_callback,
+        DeallocCallback dealloc_callback,
+        void *userdata
+    ) {
+        return new CustomAllocatorTrampoline(alloc_callback, dealloc_callback, userdata);
```
tgolsson, line 248: maybe we should have a way of destroying this? :-)

tgolsson (approved):
One comment, otherwise LGTM.

physx-sys/src/physx_api.cpp
hrydgard: We do. It's a PxAllocatorCallback, just ->Release() it.

PR #97, opened 2020-10-24
Add deps.rs and Ark Discord badges

repi:
Looks quite nice and likely should have both on all of our Rust crates.

https://deps.rs recently got revived and has new mantainers, and our Ark Discord is pretty new also


Jake-Shadle (approved):
Ohh, I should use the proper discord one on cargo-deny, I just hacked one.

PR #105, opened 2020-11-22
Fix warnings found by clippy

NickCaplinger:
One error is caused by a missing argument to panic!.
The others seem to be related to the recent glam upgrade.


repi (approved):
Thanks! 

And yes those Vec field accessor functions indeed got deprecated on glam

PR #111, opened 2020-12-03
Update glam to 0.11.1

VZout:
Just found out that `0.11` doesn't automatically gives us any 0.11 patch release. We need 0.11.1 since it includes Jasper's latest changes.

### DIFF
--- a/physx/Cargo.toml
+++ b/physx/Cargo.toml
@@ -18,7 +18,7 @@ physx-sys = { version = "0.4.10", path = "../physx-sys" }
 
 enumflags2 = "0.6"
 log = "0.4"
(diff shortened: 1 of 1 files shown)


repi (commented):
So this shouldn't be needed. It is only required if the physx library requires 0.11.1.

Otherwise it is up to the app crate that is using `physx` and its `Cargo.lock` file which patch version of glam that is used

VZout:
> So this shouldn't be needed. It is only required if the physx library requires 0.11.1.
> 
> Otherwise it is up to the app crate that is using `physx` and its `Cargo.lock` file which patch version of glam that is used

Ah okay good.

PR #113, opened 2020-12-09
Fix issues I ran into when porting our code to the new API

hrydgard:
First round of changes. Will require a major API version bump once it's all in.

Generally this wasn't too bad but there are holes.

A lot of the SceneDesc parameters were lost, added back a few. It's gonna be more though, so should we add a builder?

Now includes @Hentropy 's new object creation rework.

### DIFF
--- a/physx/src/physics.rs
+++ b/physx/src/physics.rs
@@ -105,6 +105,24 @@ pub struct PhysicsFoundation<Allocator: AllocatorCallback, Geom: Shape> {
     extensions_loaded: bool,
 }
 
+unsafe impl<T, Allocator: AllocatorCallback, Geom: Shape> Class<T>
+    for PhysicsFoundation<Allocator, Geom>
+where
+    physx_sys::PxPhysics: Class<T>,
+{
+    fn as_ptr(&self) -> *const T {
+        self.physics.obj.as_ptr()
+    }
+
+    fn as_mut_ptr(&mut self) -> *mut T {
+        self.physics.obj.as_mut_ptr()
+    }
+}
+
+impl<Allocator: AllocatorCallback, Geom: Shape> Physics for PhysicsFoundation<Allocator, Geom> {
+    type Shape = Geom;
+}
+
 impl<Allocator: AllocatorCallback, Geom: Shape> PhysicsFoundation<Allocator, Geom> {
     pub fn new(allocator: Allocator) -> PhysicsFoundation<Allocator, Geom> {
         let mut foundation =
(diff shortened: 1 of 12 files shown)


physx/src/physics.rs
```diff
@@ -100,9 +100,9 @@ pub const PX_PHYSICS_VERSION: u32 = crate::version(4, 1, 1);
 /// Parametrized by the Foundation's Allocator and the Physics' Shape type.
 pub struct PhysicsFoundation<Allocator: AllocatorCallback, Geom: Shape> {
     // Order matters here for Drop. Foundation must be dropped last.
-    pub physics: Owner<PxPhysics<Geom>>,
+    physics: Owner<PxPhysics<Geom>>,
     pub pvd: Option<VisualDebugger>,
```
tgolsson, line 104: Let's be consistent and make this one private too while we're at it.

hrydgard:
PxSceneDesc::new is really getting too many arguments, as mentioned. Won't scale. Builder or parameter struct?

tgolsson:
I'm a fan of the builder pattern but in this case I think a parameter struct would be better. Maybe just convert PxSceneDesc to that?

Hentropy:
I was thinking of a descriptor pattern for the objects created by physics.  Something like the following:
```rust
pub trait Physics {
    fn create<Desc: Descriptor>(&mut self, desc: Desc) -> Desc::Target {
        desc.create(self)
    }
// ... rest of Physics interface
}

pub trait Descriptor {
    type Target;
    fn create(&self, physics: &mut impl Physics) -> Self::Target;
}
```

I have a pretty clear idea of what to do here, I just never got around to implementing it. I'll write it up and add it to this PR.

Hentropy:
I can't add directly to this so see #114

physx/src/scene.rs
```diff
@@ -923,31 +968,89 @@ impl From<PxPairFilteringMode::Enum> for PairFilteringMode {
...
+    }
+}
+
 #[derive(Debug, Clone, Copy)]
 #[repr(u32)]
 pub enum SolverType {
-    PGS = 1,
-    TGS = 2,
+    PGS = 0,
+    TGS = 1,
```
tgolsson, line 1053: Can't we use the Px constants here directly?

tgolsson (changes requested):

tgolsson:
@hrydgard  Also please update the changelog!

Hentropy:
#117 could be rolled into this as well, I left it marked as a draft so I wasn't pinging Maik and Tomasz.

hrydgard:
Yeah, I'll roll that in later today.

physx/src/scene.rs
hrydgard: I guess we could, was just following the pattern of other enums. Don't think it really matters though.

hrydgard:
I think we can get this in as-is, and do any further fixes separately before we tag a new release.

tgolsson (approved):
LGTM

PR #114, opened 2020-12-10
add descriptors

Hentropy:
This adds a `create` method to the Physics trait that uses a `Descriptor` trait implemented on a handful of descriptor types (Scene, RigidStatic, RigidDynamic, Shape, Material, Constraint) to simplify object creation.  This is a particularly big win for the `SceneDescriptor`.  The old `create_*` methods have not been removed.  Articulation, ArticulationReducedCoordinate and Aggregate do not have descriptors since their descriptors need extra design work so that the objects that they will contain can be easily described in the descriptor (or it really isn't worth having one).


tgolsson (commented):
LGTM, I'll leave final call to @hrydgard.

physx/src/traits/descriptor.rs
```diff
@@ -0,0 +1,395 @@
...
+    pub fn new(user_data: U) -> Self {
+        Self {
+            phantom_marker: PhantomData,
+            user_data,
+            on_collide: None,
+            on_trigger: None,
+            on_constraint_break: None,
+            on_wake_sleep: None,
+            on_advance: None,
+            gravity: Default::default(),
```
hrydgard, line 119: Could we write all of these out instead of using Default::default()?   In case of gravity here, this will be zero, right? I think we should default gravity to 0, -9.81, 0, for example.

hrydgard:
This is far better, thanks! Just a question about the default values.

physx/src/traits/descriptor.rs
tgolsson: We should use the default PhysX values, in my opinion. Especially as some engines use Z-up, and some don't want gravity.

hrydgard: Oh, hm. I thought that was the default, but of course, zero is the right value here.

Anyway, for bools and enums, I don't really like to see Default::default().

Hentropy: I replaced all the defaults, and changed gravity to be -9.81 y.  PhysX default is 0g 🤷

tgolsson: I don't think we should change gravity from default; that seems like a gotcha.

Hentropy: Oops, shoulda refreshed before I commited

physx/src/scene.rs
```diff
@@ -934,31 +968,89 @@ impl From<PxPairFilteringMode::Enum> for PairFilteringMode {
...
-            BroadPhaseType::GPU => 3,
+            BroadPhaseType::SweepAndPrune => PxBroadPhaseType::eSAP,
+            BroadPhaseType::MultiBoxPruning => PxBroadPhaseType::eMBP,
+            BroadPhaseType::AutomaticBoxPruning => PxBroadPhaseType::eABP,
+            BroadPhaseType::GPU => PxBroadPhaseType::eGPU,
         }
     }
 }
 
+impl Default for BroadPhaseType {
```
hrydgard, line 1043: Does this kind of Default for a type even make sense? I would normally think that defaults should be specified in the default constructor of the type whose member is a BroadPhaseType in this case (basically, what I just had you change)...

I'm OK with having it though but it's unintuitive to me.

hrydgard (approved):

PR #117, opened 2020-12-11
impl Physics for PhysicsFoundation update readmes

Hentropy:
Implementing `Physics` for `PhysicsFoundation` makes the API feel much cleaner, calling a method to retrieve the `Physics` object was annoying.  `Foundation` and `Pvd` could be implemented as well, but they aren't used as frequently.

This also updates the readmes code examples, explanations and alternatives list.

builds on #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@ Rust binding and wrapper over [NVIDIA PhysX](https://github.com/NVIDIAGameWorks/
 
 Created and maintained by [Embark](http://embark.games) and _**not**_ officially supported by NVIDIA.
 
-This repository contains 3 crates:
+This repository contains 2 crates:
 
 | Name | Description | Links |
 | --- | --- | --- |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -49,14 +51,13 @@ The following code example shows how [`physx`](physx/) can be initialized.
 const PX_PHYSICS_VERSION: u32 = physx::version(4, 1, 1);
 let mut foundation = Foundation::new(PX_PHYSICS_VERSION);
 
-let mut physics = PhysicsBuilder::default()
-    .load_extensions(false) // switch this flag to load extensions during setup
```
hrydgard, line 53: Hm, where is this flag set now?

Hentropy: There is a `PhysicsFoundationBuilder` which has a `with_extensions` method.

hrydgard:
I merged this into #113.

Issue #99, opened 2020-10-27
Add trampolines for PxSweepCallback, PxRaycastCallback and PxOverlapCallback


Hentropy:
#### The Problem
Currently, scene queries are not usable because the callback types they need are not instantiable.

#### The Solution
Add trampolines for PxSweepCallback, PxRaycastCallback and PxOverlapCallback.

#### Alternatives
I tried passing in the callback viz MaybeUninit, but it crashes with a memory access violation.  It's theoretically possible to cobble the vtable together in Rust, but a trampoline is the best solution.

#### Context
I have written [call throughs](https://github.com/Hentropy/physx-rs/blob/main/physx/src/scene.rs#L502) for `sweep`, `overlap`, and `raycast` but without these types they cannot be called.

NickCaplinger:
Hey @Hentropy, I've got a rough implementation working for this right now that supports usage of PxRaycastBuffer and user-defined PxRaycastCallback implementations. I should have a PR ready this weekend for raycasts, sweeps, and overlaps and I'd love to have you review it when you have the chance. :)

Hentropy:
@NickCaplinger You're a legend, thank you.  Gimme a ping and I'll take a look 😄

Jake-Shadle:
Closed in #181

PR #119, opened 2020-12-13
Add initial support for scene raycast, sweep, and overlap

NickCaplinger:
~Note that this pull request also reorganizes the examples, putting them into their own crate. I used [ash](https://github.com/MaikKlein/ash/tree/master/examples) as my example. This is likely going to break the current Github Actions, so I'll need to update that as well.~ Nevermind, I didn't actually know that Cargo had an explicit "examples" feature. Neat! Reverting these changes.

Simple hit buffers and full custom user-defined callbacks are supported.
It's not currently possible to specify the size of the buffer given to user-defined callbacks, but if everything else looks good, we can add support for that too.

Addresses https://github.com/EmbarkStudios/physx-rs/issues/99

@Hentropy

### DIFF
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -285,16 +285,22 @@ extern "C" {
     pub fn create_raycast_callback(
         process_touches_callback: RaycastProcessTouchesCallback,
         finalize_query_callback: FinalizeQueryCallback,
+        touches_buffer: *mut PxRaycastHit,
+        num_touches: u32,
         userdata: *mut c_void,
     ) -> *mut PxRaycastCallback;
     pub fn create_sweep_callback(
         process_touches_callback: SweepProcessTouchesCallback,
         finalize_query_callback: FinalizeQueryCallback,
+        touches_buffer: *mut PxSweepHit,
+        num_touches: u32,
         userdata: *mut c_void,
     ) -> *mut PxSweepCallback;
     pub fn create_overlap_callback(
         process_touches_callback: OverlapProcessTouchesCallback,
         finalize_query_callback: FinalizeQueryCallback,
+        touches_buffer: *mut PxOverlapHit,
+        num_touches: u32,
         userdata: *mut c_void,
     ) -> *mut PxOverlapCallback;
 
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -241,7 +241,18 @@ pub struct FilterShaderCallbackInfo {
     pub constantBlockSize: u32,
 }
 
+pub type PxAgain = bool;
+
 pub type SimulationFilterShader = unsafe extern "C" fn(*mut FilterShaderCallbackInfo) -> u16;
+
+pub type RaycastProcessTouchesCallback =
+    unsafe extern "C" fn(*const PxRaycastHit, u32, *const c_void) -> PxAgain;
+pub type SweepProcessTouchesCallback =
+    unsafe extern "C" fn(*const PxSweepHit, u32, *const c_void) -> PxAgain;
+pub type OverlapProcessTouchesCallback =
+    unsafe extern "C" fn(*const PxOverlapHit, u32, *const c_void) -> PxAgain;
+
+pub type FinalizeQueryCallback = unsafe extern "C" fn(*const c_void);
 
 pub type AllocCallback =
     unsafe extern "C" fn(u64, *const c_void, *const c_void, u32, *const c_void) -> *mut c_void;
@@ -268,6 +279,26 @@ extern "C" {
         callback: RaycastHitCallback,
         userdata: *mut c_void,
     ) -> *mut PxQueryFilterCallback;
+
+    pub fn create_raycast_buffer() -> *mut PxRaycastCallback;
+    pub fn create_sweep_buffer() -> *mut PxSweepCallback;
+    pub fn create_overlap_buffer() -> *mut PxOverlapCallback;
+
+    pub fn create_raycast_callback(
+        process_touches_callback: RaycastProcessTouchesCallback,
+        finalize_query_callback: FinalizeQueryCallback,
+        userdata: *mut c_void,
+    ) -> *mut PxRaycastCallback;
+    pub fn create_sweep_callback(
+        process_touches_callback: SweepProcessTouchesCallback,
+        finalize_query_callback: FinalizeQueryCallback,
+        userdata: *mut c_void,
+    ) -> *mut PxSweepCallback;
+    pub fn create_overlap_callback(
+        process_touches_callback: OverlapProcessTouchesCallback,
+        finalize_query_callback: FinalizeQueryCallback,
+        userdata: *mut c_void,
+    ) -> *mut PxOverlapCallback;
 
     pub fn create_alloc_callback(
         alloc_callback: AllocCallback,
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -305,6 +305,8 @@ extern "C" {
     ) -> *mut PxOverlapCallback;
 
     pub fn delete_raycast_callback(callback: *mut PxRaycastCallback);
+    pub fn delete_sweep_callback(callback: *mut PxSweepCallback);
+    pub fn delete_overlap_callback(callback: *mut PxOverlapCallback);
 
     pub fn create_alloc_callback(
         alloc_callback: AllocCallback,
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -241,16 +241,14 @@ pub struct FilterShaderCallbackInfo {
     pub constantBlockSize: u32,
 }
 
-pub type PxAgain = bool;
-
 pub type SimulationFilterShader = unsafe extern "C" fn(*mut FilterShaderCallbackInfo) -> u16;
 
 pub type RaycastProcessTouchesCallback =
-    unsafe extern "C" fn(*const PxRaycastHit, u32, *const c_void) -> PxAgain;
+    unsafe extern "C" fn(*const PxRaycastHit, u32, *const c_void) -> bool;
 pub type SweepProcessTouchesCallback =
-    unsafe extern "C" fn(*const PxSweepHit, u32, *const c_void) -> PxAgain;
+    unsafe extern "C" fn(*const PxSweepHit, u32, *const c_void) -> bool;
 pub type OverlapProcessTouchesCallback =
-    unsafe extern "C" fn(*const PxOverlapHit, u32, *const c_void) -> PxAgain;
+    unsafe extern "C" fn(*const PxOverlapHit, u32, *const c_void) -> bool;
 
 pub type FinalizeQueryCallback = unsafe extern "C" fn(*const c_void);
 
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -304,6 +304,8 @@ extern "C" {
         userdata: *mut c_void,
     ) -> *mut PxOverlapCallback;
 
+    pub fn delete_raycast_callback(callback: *mut PxRaycastCallback);
+
     pub fn create_alloc_callback(
         alloc_callback: AllocCallback,
         dealloc_callback: DeallocCallback,
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -244,13 +244,13 @@ pub struct FilterShaderCallbackInfo {
 pub type SimulationFilterShader = unsafe extern "C" fn(*mut FilterShaderCallbackInfo) -> u16;
 
 pub type RaycastProcessTouchesCallback =
-    unsafe extern "C" fn(*const PxRaycastHit, u32, *const c_void) -> bool;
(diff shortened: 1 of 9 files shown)


physx/examples/scene_queries.rs
```diff
@@ -0,0 +1,258 @@
...
+        // Overlaps
+        let mut overlap_callback = PxOverlapBuffer::new();
+        let did_overlap = scene.overlap(
+            &sphere_geo,
+            &PxTransform::default(),
+            &mut overlap_callback,
+            &filter_data,
+            None,
+        );
+        assert!(did_overlap); // overlaps with ground plane
```
Hentropy, line 218: This assertion fails after getting this error from PhysX: `..\physx-rs\physx-sys\PhysX/physx\source/physx/src\NpSceneQueries.cpp (709) : invalid parameter : PxScene::overlap() and PxBatchQuery::overlap() calls without eANY_HIT flag require a touch hit buffer for return results.`
If I comment this assertion out, I get the same error again for the other overlap call, but sweep and raycast work great.

physx-sys/src/lib.rs
```diff
@@ -241,8 +241,19 @@ pub struct FilterShaderCallbackInfo {
     pub constantBlockSize: u32,
 }
 
+pub type PxAgain = bool;
```
Hentropy, line 244: I think just using a `bool` is fine.  The traits that expose the functions that return this just return a `bool`, so the end-user currently isn't seeing this at all.

Hentropy (changes requested):
Hey, this is looking good.  Couple of issues to address, any comments on one hit type apply to the other two as well.

physx/src/query.rs
```diff
@@ -0,0 +1,290 @@
...
+    ) -> bool {
+        // Unused by this impl
+        false
+    }
+
+    unsafe extern "C" fn finalize_query(_user_data: *const c_void) {
+        // Unused by this impl
+    }
+
+    unsafe fn into_px(&mut self) -> *mut physx_sys::PxSweepCallback {
```
tgolsson, line 176: semantically, I think this should be `as_raw` or `as_mut_ptr`? `into` is often used for sinking calls, which this isn't.

physx/src/scene.rs
```diff
@@ -630,14 +633,135 @@ pub trait Scene: Class<physx_sys::PxScene> + UserData {
...
+        filter_data: &PxQueryFilterData,
+        filter_callback: Option<&mut PxQueryFilterCallback>,
+        cache: Option<&PxQueryCache>,
+        inflation: f32,
+    ) -> bool {
+        let filter_callback = filter_callback
+            .map_or(std::ptr::null_mut::<PxQueryFilterCallback>(), |v| {
+                v as *mut _
+            });
+        let cache = cache.map_or(std::ptr::null::<PxQueryCache>(), |v| v as *const _);
```
tgolsson, line 713: I personally find `.map(...).unwrap_or(...)` clearer. Not a required change, but it's an idiom I see a lot more in the wild as opposed to `map_or`.

physx/src/query.rs
```diff
@@ -0,0 +1,290 @@
+// Author: Tom Olsson <tom.olsson@embark-studios.com>
+// Copyright © 2020, Embark Studios, all rights reserved.
+// Created: 9 December 2020
```
tgolsson, line 3: I don't think I'm the author of this file :-)

physx/src/query.rs
```diff
@@ -0,0 +1,290 @@
...
+
+//! Scene queries are used to perform tests against the actors and geometry held in a scene.
+
+#![warn(clippy::all)]
+#![allow(clippy::missing_safety_doc)]
+#![allow(deprecated)]
```
tgolsson, line 9: are all these required? :-)

physx/src/query.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    unsafe fn into_px(&mut self) -> *mut physx_sys::PxRaycastCallback {
+        self.0.obj.as_mut_ptr()
+    }
+}
+
+/// A trait for creating custom raycast callbacks for PhysX.
+pub unsafe trait RaycastCallback: Sized {
+    unsafe fn new() -> Self;
+    unsafe fn get_raycast_hit(&mut self) -> Option<physx_sys::PxRaycastHit>;
```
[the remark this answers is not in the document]
tgolsson: Very much agreed; I think this'd also make it possible to remove the `Sized` bound when combined with your other suggestion.

physx/src/scene.rs
```diff
@@ -630,14 +633,135 @@ pub trait Scene: Class<physx_sys::PxScene> + UserData {
...
+    /// Casts a ray into the scene, reporting any hit objects via the [`RaycastCallback`] trait passed into `hit_callback`.
+    /// For simple usage, passing in [`PxRaycastBuffer`] will suffice.
+    /// Users can also provide a custom implementation of [`RaycastCallback`] in cases where allocating a single buffer to store
+    /// hits is prohibitively expensive.
+    /// Returns true if the ray hit an actor.
+    /// See the [PhysX documentation on `raycast`][src] for more details.
+    ///
+    /// [src]: (https://gameworksdocs.nvidia.com/PhysX/4.1/documentation/physxapi/files/classPxScene.html#a7d7dcd877cee092f8b57c67d79982b50)
+    #[allow(clippy::too_many_arguments)]
+    fn raycast(
```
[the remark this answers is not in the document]
tgolsson: Agreed; though I think this'd be a "SceneQueryExt" trait if so; and we'd still supply the "raw" mapping. But I often end up doing this kind of wrapping myself both as a small cache with the query struct and ergonomics.

tgolsson (commented):
Mostly agree with @Hentropy, some minor comments and thoughts I've added.

physx/src/query.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    /// # Safety
+    /// Do not override this method.
+    unsafe fn into_px(&mut self) -> *mut physx_sys::PxRaycastCallback {
+        create_raycast_callback(
+            Self::process_touches,
+            Self::finalize_query,
+            Box::into_raw(Box::new(self)) as *mut c_void,
+        )
+    }
```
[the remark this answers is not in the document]
Hentropy: Looking at this again, I don't think that `into_px` function is correct, `Scene::raycast` is creating a `physx_sys::PxRaycastCallback` every time it's called.  The `user_data` also needs to be cleaned up.
The starting point is to make a new-type wrapper with a generic, but it's gonna get a little tricky.
```Rust
pub struct PxRaycastCallback<T: RaycastCallback> {
    obj: physx_sys::PxRaycastCallback,
    phantom_user_data: PhantomData<T>,
}

impl<T: RaycastCallback> PxRaycastCallback<T> {
    fn new(user_data: T) -> Option<Owner<Self>> {
        create_raycast_callback(
            Self::process_touches,
            Self::finalize_query,
            Bow::into_raw(Box::new(user_data)), // we need to make sure this is dropped
        )
    }

    unsafe extern "C" fn process_touches(
        buffer: *const physx_sys::PxOverlapHit,
        num_hits: u32,
        user_data: *const c_void,
    ) -> bool {
        (&*(user_data as *const T)).process_callbacks(slice::from_raw(buffer, num_hits as usize))
    }
    // finalize_query wrapper as well
}
```
The drop impl will need a `user_data` getter and a dtor for the rest of it in C++, and that will need to be integrated in a way that doesn't allow calling something like `PxRaycastBuffer::get_user_data(...)` because it lacks the trampoline fields.  `into_px` should consume `self` and produce `PxRaycastCallback<Self>`.  Some methods/traits may need to be shifted around so that there's a sensible and safe trait bound exposed in `Scene::raycast`.

User data makes everything messy 😢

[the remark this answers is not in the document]
Hentropy: `physx_sys::PxRaycastHit` (and co.) should also get safe wrappers, that could be done in a follow up though.

NickCaplinger:
Thank you two so much for the thorough review! I'll address these points tonight after work for my day job is done. Cheers. :)

NickCaplinger:
@tgolsson @Hentropy Hey guys! It's been a busy week, but I managed to find time to work on this.
I've left these forced-pushed commits unsquashed, but let me know if it's easier to review a squashed commit and I'll make that change.

I wasn't very satisfied with the initial solution and reworked the main traits using some of the feedback you guys gave.
The traits are now safe to use and the user callbacks from C++ into Rust are wrapped with catch_unwind to prevent unwinding across the FFI barrier. C++ callback objects are now dropped correctly and deleted on the C++ side. Hit buffers are exposed to the end user.

There are still some issues with it, but I'd love to hear feedback about the direction of this. I'll drive this PR to completion when I'll have plenty of free time this weekend! Thanks for your patience - this project has been a great Rust learning experience for me. :)

Some thoughts on work remaining:
* Queries are still not very idiomatic and more C++ism leaks through than I'd like in the form of touch buffer management.
* I don't like that RaycastCallbackDefault et al. exist but I'm not sure how best to deal with that problem. Perhaps a better, less verbose name would be enough.
* SceneQueryExt is not yet implemented, but I love the idea.
* The Px{Raycast,Sweep,Overlap}Hit structs are not wrapped yet. I suspect wrapping these will greatly improve usability for very little effort.

physx/src/query.rs
```diff
@@ -40,251 +38,374 @@ impl Into<physx_sys::PxQueryFilterData> for PxQueryFilterData {
...
 
-#[repr(transparent)]
-pub struct PxRaycastCallback {
-    pub obj: Owner<physx_sys::PxRaycastCallback>,
-}
+// RAYCAST CALLBACKS
 
-#[repr(transparent)]
-pub struct PxRaycastBuffer(PxRaycastCallback);
+pub struct RaycastCallbackDefault;
```
Hentropy, line 43: The use-case here should be filled by a `PxRaycastBuffer` wrapper instead.  There should be a `RaycastCallback` trait for the bound used in `Scene::query` for consistency with the class-as-trait pattern.  However, I'm not sure if the current `RaycastCallback` trait will work right now, you may need `Class<physx_sys::PxRaycastCallback>` and maybe `UserData` (see the other note about that, `PxRaycastBuffer` would define it as `()`) as super traits.

Also, not sure whether using `*Callback` as the name for user defined trait overrides or `*/Px*` as class `trait/struct` names should take precedence here. It might be possible to overlap the concerns, but it's kinda confusing me trying to describe it here so perhaps they should get split.

physx-sys/src/physx_api.cpp
```diff
@@ -294,27 +350,53 @@ extern "C"
...
     ) {
-		return new HitCallbackTrampoline<PxRaycastHit>(process_touches_callback, finalize_query_callback, userdata);
+        return new RaycastHitCallbackTrampoline(
+            process_touches_callback, finalize_query_callback, touchesBuffer, numTouches, userdata);
+    }
+
+    void delete_raycast_callback(PxRaycastCallback *callback)
+    {
+        delete callback;
+    }
```
Hentropy, line 366: Should be a `get_raycast_callback_user_data` here, otherwise it's irretrievable from Rust and it'll get leaked on drop.

Hentropy (commented):
Looking good!  The error from CI is because of the direct struct constructor because Linux and Windows have different struct layouts for some things (yay).

Now that the templates are gone, this is shaping up more like the class as trait/struct pattern of `RigidDynamic/PxRigidDynamic` than the class as user defined trait pattern of `SimulationEventCallback` I was thinking it would resemble.  After telling you to remove that second `Class<T>` impl ... you'll probably want it back again (sorry).

NickCaplinger:
Thanks for the review again! I didn't get around to working on this last weekend like I thought I would - I got a new GPU on Thursday, was finally able to play Cyberpunk, and underestimated how much I'd want to keep playing it. 😱 

> Looking good! The error from CI is because of the direct struct constructor because Linux and Windows have different struct layouts for some things (yay).

Oops, good call. I suppose that's another argument for implementing that wrapper. :)

> Now that the templates are gone, this is shaping up more like the class as trait/struct pattern of `RigidDynamic/PxRigidDynamic` than the class as user defined trait pattern of `SimulationEventCallback` I was thinking it would resemble. After telling you to remove that second `Class<T>` impl ... you'll probably want it back again (sorry).

True! That does make sense, looking back at the existing patterns.

Hentropy:
@NickCaplinger ping to keep this on your radar. It's looking really good, and it's very nearly there 🤗

rlidwka:
What is the current status of this pull request?

NickCaplinger:
Hi @rlidwka and @Hentropy (and sorry for the radio silence!).

I took a very long hiatus away from games around this time, so what you see is what you get here, unfortunately. I haven't been following the project since then. @Hentropy might know, though. I'm not sure what to do with this PR until then.

Thank you for the positive words of encouragement by the way, Henno. I never replied, but it was helpful in other parts of my life at the time.

Hentropy:
I'm glad to hear that Nick ☺️

@rlidwka I'm in the same boat as Nick here, sorry. You could pick it up where it's been loft off, though I'm not sure how much the rest of the library has changed since this was last updated.

Jake-Shadle:
#181 added support for the low level bindings, so feel it's better to close this and a future PR can add the friendly wrappers to the physx crate in a smaller PR.

PR #120, opened 2020-12-14
Make convex and triangle mesh validation optional

NickCaplinger:
This change makes triangle mesh validation optional, allowing for potential performance improvements when the mesh has already been validated. The API user is still responsible for validating their mesh.

I've replaced the mesh validation with mesh descriptor validation.


hrydgard (approved):

hrydgard:
Oh yeah, I guess this was missing changelog and a comment. I'll take care of that when I do the next release.

NickCaplinger:
Thank you @hrydgard!

Issue #92, opened 2020-10-17
Add support for aarch64-apple-darwin


repi:
We'll need to add support for the new Mac on Apple Silicon target `aarch64-apple-darwin`.

The target just became Tier 2 in Rust and is part of the nightly builds as of a few days ago (https://github.com/rust-lang/rust/issues/73908#issuecomment-710101535) so one can test building both on the real DTK preview hardware as well as cross-compile from Mac.

Currently when building one gets this: `thread 'main' panicked at 'unknown TARGET triple 'aarch64-apple-darwin''`

repi:
This is even more relevant for us now as the hardware is generally available next week!

PR #123, opened 2020-12-19
Initial aarch64-apple-darwin support

repi:
Adds initial support for Mac M1 ARM CPUs.

This may not be correct as had problems running our pxbind structgen utility on Mac M1 so for now just copy'n'pasted the x86_64 Mac structgen files. Does seem to run but could be subtle differences (that could cause catastrophic failures)?

Resolves: #92

### DIFF
--- a/physx-sys/build.rs
+++ b/physx-sys/build.rs
@@ -155,7 +155,7 @@ fn main() {
         let mut include = PathBuf::from("src/generated");
 
         match target.as_str() {
-            "x86_64-apple-darwin" | "x86_64-pc-windows-msvc" | "aarch64-linux-android" => {
+            "x86_64-apple-darwin" | "x86_64-pc-windows-msvc" | "aarch64-linux-android" | "aarch64-apple-darwin" => {
                 include.push(target);
             }
             nix if nix.starts_with("x86_64-unknown-linux") => {
--- a/physx-sys/build.rs
+++ b/physx-sys/build.rs
@@ -155,7 +155,10 @@ fn main() {
         let mut include = PathBuf::from("src/generated");
 
         match target.as_str() {
-            "x86_64-apple-darwin" | "x86_64-pc-windows-msvc" | "aarch64-linux-android" | "aarch64-apple-darwin" => {
+            "x86_64-apple-darwin"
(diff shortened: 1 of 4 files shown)


hrydgard:
I'm not actually sure what would be different between archs here. It's just a bunch of structs with repr(C), and all the platforms we support are 64-bit and, I think, pretty much entirely layouts structs the same way? Do we really need platform specific structgen.rs ?

repi:
Yeah I'm not fully sure either, I've done it this way before also of just copying it over from other platforms, @h3r2tic do you know more specifically where compiler/arch differences could affect the structgen?

repi:
@h3r2tic what do you think of the risks of just getting this in vs generating a proper definition?

Seems to be working well, been using it for a couple of months and no seen physx issues

tgolsson:
I spent a couple of evenings trying to replace the bindgen stuff with an easier to maintain Python version but the clang-bindings lacked some things, so getting a 1-1 match in output wasn't possible.

tgolsson:
Added an issue here: #127, if we want to tackle it.

h3r2tic:
> @h3r2tic what do you think of the risks of just getting this in vs generating a proper definition?
> 
> Seems to be working well, been using it for a couple of months and no seen physx issues

In reality: who knows! If we reuse x64 struct definitions for arm64, we are crossing fingers and relying on C++ ABI compatibility between those (we rely on repr(C) between C and Rust, but the generated struct layout matches the C++ ABI). In theory we shouldn't be relying on pre-generated struct layouts at all, although they are de-facto stable.

So we _should_ run structgen on aarch64-apple-darwin, but the build process has changed for clang/libtooling, and I'm at a loss trying to get it to work. I'd much rather go with @tgolsson's idea of relying on the devs of Python's or Rust's bindings, and using a sane build system.

Seeing as that's not something we have right now, and considering you've been testing aarch64-apple-darwin for a while, we could just merge this, and _at some point_ replace the generator with one written in _not-C++_.

repi:
Thanks! and agreed, think we should get this in @hrydgard @tgolsson

hrydgard (approved):

tgolsson (approved):

PR #126, opened 2021-02-18
Release physx 0.11.0

### LINKED ISSUE #125: New patch release?
With #124 merged, it would nice to have it released so that we can use `glam` 0.12 in Ark.

XAMPPRocky:
resolves #125

### DIFF
--- a/physx/CHANGELOG.md
+++ b/physx/CHANGELOG.md
@@ -103,8 +103,8 @@
 - Ability to not run the default filter shader before the callback.
 - Fix for triangle mesh data when using glam with SSE enabled
 
-[Unreleased]: https://github.com/EmbarkStudios/physx-rs/compare/physx-v0.10.0...HEAD
-[0.10.0]: https://github.com/EmbarkStudios/physx-rs/compare/physx-v0.9.0...physx-v0.10.0
+[Unreleased]: https://github.com/EmbarkStudios/physx-rs/compare/physx-v0.11.0...HEAD
+[0.10.0]: https://github.com/EmbarkStudios/physx-rs/compare/physx-v0.9.0...physx-v0.11.0
 [0.9.0]: https://github.com/EmbarkStudios/physx-rs/compare/physx-v0.8.0...physx-v0.9.0
 [0.8.0]: https://github.com/EmbarkStudios/physx-rs/compare/physx-v0.7.4...physx-v0.8.0
 [0.7.4]: https://github.com/EmbarkStudios/physx-rs/compare/physx-v0.7.3...physx-v0.7.4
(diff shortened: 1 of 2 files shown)


hrydgard (approved):

physx-sys/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "physx-sys"
 description = "Unsafe bindings for NVIDIA PhysX C++ SDK"
-version = "0.4.11"
+version = "0.11.0"
```
hrydgard, line 4: Wait, no, we usually don't bump the sys crate if we just bump the physx crate. They don't have synced versions.

(should they, though?)

tgolsson: I don't think we should synchronize the versions, that not be correct semantic versioning. Also, is this really a minor version bump or just a patch?

XAMPPRocky: I'll revert that for now, but generally I think they should be synced. It just makes it easier to maintain in my opinion, and it's more intuitive for users. For example imagine the case where someone is using both physx and physx-sys in a project, how would they know which version of the sys crate is compatible with the high level crate? When they're the same version it's pretty intuitive, it's just the same version.

XAMPPRocky: > I don't think we should synchronize the versions, that not be correct semantic versioning. Also, is this really a minor version bump or just a patch?

It's a minor bump, because glam is used in the public API, and if you are also using glam, then suddenly you can't use `cargo update` because `0.10.0` would have glam `0.11.x` and `0.10.1` would have glam `0.12`. Which means that at best you have a duplicate dependency, and at worst your project no longer compiles until you also upgrade to glam `0.12`.

hrydgard (commented):
Please also update the diff links at the bottom of the changelog, similarly to this commit: https://github.com/EmbarkStudios/physx-rs/commit/b56e585e36d877d8d7ad08f9746d344cfb10a2c6

Sorry I missed commenting on that before.

physx-sys/Cargo.toml
tgolsson: Makes sense regarding glam. 

In that specific case for sys + high-level it's more intuitive, but physx already lists physx-sys as a dependency. If we expect users to work with both versions of the crates we should use the version field there + maybe in the README to document the relation.

repi: we indeed do not sync the versions, only bump sys version when there is an update in that crate or actual physx upgrade

hrydgard (approved):

repi (approved):

tgolsson (approved):

PR #128, opened 2021-03-06
Use Embark standard lints v0.2

repi:
Switches to use our new standard template for defining shared Rust and Clippy lints for our Embark projects, https://github.com/EmbarkStudios/rust-ecosystem/issues/59.

Did a couple of lint fixes from this but didn't want to make the change too intrusive so added some allow exceptions for now that can be revisited


tgolsson (commented):
@repi I'm fine with merging this, but who is responsible for keeping this updated and making relevant changes if our standard changes?

hrydgard (approved):
Approving with the same thought as Tom.

repi:
@tgolsson @hrydgard think it should be part of regular maintenance by primarily the codeowners of the crates. But after this is merged it is not an urgent thing to upgrade to the latest lint collection rapidly, one can do when one is doing other changes similar to updating dependencies and such.

And a few of us will probably do some drive-by updates like this and submit PRs occasionally to test out new lints and fix up code, but think the main responsibility should be on the codeowners for the larger crates, like every few months or so or when doing other larger changes or maintenance.

PR #129, opened 2021-03-29
Make glam an optional dependency

XAMPPRocky:
Currently upgrading rust-gpu in Ark is usually blocked on physx-rs because Ark disallows duplicate deps and glam's minor version number changes often. This PR changes it so `glam` is optional dependency of `physx`, which should allow Ark to not depend on `glam` from `physx`.

### DIFF
--- a/physx/Cargo.toml
+++ b/physx/Cargo.toml
@@ -18,7 +18,7 @@ physx-sys = { version = "0.4.10", path = "../physx-sys" }
 
 enumflags2 = "0.6"
 log = "0.4"
-glam = "0.12"
+glam = { version = "0.13", optional = true }
 thiserror = "1.0.20"
 
 [features]
--- a/physx/CHANGELOG.md
+++ b/physx/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## [Unreleased]
 
+- [PR#129](https://github.com/EmbarkStudios/physx-rs/pull/129) Glam is now an
+  optional dependency and has been updated. You can enable the `glam` feature
+  to continue to use `glam` with `physx-sys` types.
 - [Updated various struct names and trait implementations](https://github.com/EmbarkStudios/physx-rs/pull/130):
   - `BVHStructure` is now named `BvhStructure`.
   - `PxBVHStructureDesc` is now named `PxBvhStructureDesc`.
--- a/physx/src/math/glam.rs
+++ b/physx/src/math/glam.rs
@@ -1,0 +1,55 @@
+use super::*;
+use ::glam::*;
+
+impl From<Vec3> for PxExtendedVec3 {
+    fn from(vec: Vec3) -> Self {
+        Self {
+            obj: physx_sys::PxExtendedVec3 {
+                x: vec.x as f64,
+                y: vec.y as f64,
+                z: vec.z as f64,
+            },
+        }
+    }
+}
+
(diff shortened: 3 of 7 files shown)


tgolsson:
@XAMPPRocky Thanks for the PR. I'm contemplating whether the glam feature should be on by default. And if potentially all of these glam support conversions should be in one place instead of spread throughout the code.

XAMPPRocky:
> Thanks for the PR. I'm contemplating whether the glam feature should be on by default. And if potentially all of these glam support conversions should be in one place instead of spread throughout the code.

Having it all in a single module makes sense to me, I was just trying to make the PR as low impact, I've pushed up a new version that moves all of the implementations into a single module in `math`.

As for having it be on by default, I don't think it should, as then if you need to disable it, you and all your sub-dependencies need to have it disabled in order to remove it from the dependency graph, where as being opt-in allows you to bring in the feature when you specifically need it. There's a good section in the cargo book about [default features](https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature).

hrydgard (approved):

tgolsson:
> > Thanks for the PR. I'm contemplating whether the glam feature should be on by default. And if potentially all of these glam support conversions should be in one place instead of spread throughout the code.
> 
> Having it all in a single module makes sense to me, I was just trying to make the PR as low impact, I've pushed up a new version that moves all of the implementations into a single module in `math`.
> 
> As for having it be on by default, I don't think it should, as then if you need to disable it, you and all your sub-dependencies need to have it disabled in order to remove it from the dependency graph, where as being opt-in allows you to bring in the feature when you specifically need it. There's a good section in the cargo book about [default features](https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature).

Then this constitutes a major version bump, doesn't it? I'm fine with that, but maybe it should be called out the changelog that in order to maintain the current functionality this feature has to be opted in to.

physx/Cargo.toml
```diff
@@ -18,7 +18,7 @@ physx-sys = { version = "0.4.10", path = "../physx-sys" }
 
 enumflags2 = "0.6"
 log = "0.4"
-glam = "0.12"
+glam = { version = "0.13", optional = true }
```
repi, line 21: One could also support a range of glam releases here, and update that more frequently. It is harder to test though and CI should be on the oldest version. But it is another parallel path forward as well

XAMPPRocky:
> Then this constitutes a major version bump, doesn't it? I'm fine with that, but maybe it should be called out the changelog that in order to maintain the current functionality this feature has to be opted in to.

I've updated the CHANGELOG text to be more reflective of that.

tgolsson:
@XAMPPRocky Will you do the work to fix the lints? It's not related to the work; but I don't quite have the time to do it myself right now.

tgolsson:
@XAMPPRocky If you rebase this one now it should be unblocked!

XAMPPRocky:
Rebased and updated

Jake-Shadle:
Does this mean we are going to do an actual release so don't need the git patch? (asking because the rebase broke the revision specifier we have on the patch)

XAMPPRocky:
> Does this mean we are going to do an actual release so don't need the git patch? (asking because the rebase broke the revision specifier we have on the patch)

I'll update it when I update the rust-gpu dependency.

PR #135, opened 2021-05-03
Update PhysX submodule with ARM64 buildfix for Xcode 14.5

hrydgard:
Just applies this simple fix, where inlining could cause bad intrinsics usage on a path that could not be taken (the function was called with index=2, which is not a valid parameter for that intrinsic).

https://github.com/EmbarkStudios/PhysX/commit/dc5059fb1acbf6a710c50f538b727a851c2d1139

### DIFF
--- a/physx-sys/CHANGELOG.md
+++ b/physx-sys/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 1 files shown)


tgolsson (approved):
Worth calling out in changelog maybe? I prefer more verbose than less verbose.

hrydgard:
Right, definitely belongs in the changelog. Added it.

PR #146, opened 2022-02-04
Upgrade glam from 0.13 to 0.20 (optional dependency)

R3DP1XL:
The version of glam is quite outdated (~11 mo), and makes it difficult to use conversions for those that use the optional "glam" feature.


hrydgard (approved):

hrydgard:
Just noting that I made a new release, 0.12.2, that includes this change.

PR #151, opened 2022-07-13
Remove henrik as a CODEOWNER

Jake-Shadle:


tgolsson (approved):
LGTM but CI looks rotten. You looking at that or want me to take a peek?

tgolsson:
I think just setting the AR to llvm-ar instead of the target-specific one should fix it.

PR #153, opened 2022-08-03
Update supported clang/llvm for pxbind

Nehliin:
This makes it a bit easier to add new targets since it's possible to use newer clang/llvm libs. Also updated documentation.

However it seems some of the generated bindings got some new types (did they not the currently checked in Physx version?) which causes some problems since I have some trouble re-generating all the platform specific bindings. I can do it for my own platform ofc but would be nice with some help for the rest of the platforms.


Nehliin:
ah perhps the ci will fix the problem of generating new bindings?

PR #157, opened 2022-08-05
Remove cmake support

Nehliin:
Maintenance burden on us which isn't justified since neither we nor literately anyone else seems to be using it


repi (approved):
nice!

Do add it to the CHANGELOGs also

Jake-Shadle (approved):

tgolsson (approved):
Nice! Thank you.

Please make one more PR with a small changelog note about aarch64 support that you added in the previous one.

PR #159, opened 2022-08-05
Update changelogs

Nehliin:
(assuming #158 is merged)

Not sure exactly how much of the `physx-sys` changes bleed through to `physx` so feel free to correct me if more things are needed in the `physx` changelog


tgolsson (approved):
Very nice! We *generally* don't specify what happens the actual bindings; but maybe makes sense now since they were out of sync with actual physx.

PR #167, opened 2022-10-04
Add support for setting up the physx error callback

jnises:
Oh man. c++/rust ffi is not easy 😓 
I did the stateful error callback in the end. Seemed like the cleanest interface.
Is this an ok way to do it?

### DIFF
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -354,8 +354,6 @@ pub type ZoneEndCallback = unsafe extern "C" fn(*const c_void, *const i8, bool, 
 pub type ErrorCallback =
     unsafe extern "C" fn(u32, *const c_void, *const c_void, u32, *const c_void);
 
-pub type ErrorUserdataDeleteCallback = unsafe extern "C" fn(*mut c_void);
-
 extern "C" {
     pub fn physx_create_foundation() -> *mut PxFoundation;
     pub fn physx_create_foundation_with_alloc(
@@ -365,7 +363,6 @@ extern "C" {
 
     pub fn get_default_allocator() -> *mut PxDefaultAllocator;
     pub fn get_default_error_callback() -> *mut PxDefaultErrorCallback;
-    pub fn create_default_error_callback() -> *mut PxDefaultErrorCallback;
 
     /// Destroy the returned callback object using PxQueryFilterCallback_delete.
     pub fn create_raycast_filter_callback(
@@ -395,7 +392,6 @@ extern "C" {
     pub fn create_error_callback(
         error_callback: ErrorCallback,
         userdata: *mut c_void,
-        userdata_deleter: ErrorUserdataDeleteCallback,
     ) -> *mut PxErrorCallback;
 
     pub fn destroy_error_callback(error_callback: *mut PxErrorCallback);
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -352,7 +352,7 @@ pub type ZoneStartCallback =
 pub type ZoneEndCallback = unsafe extern "C" fn(*const c_void, *const i8, bool, u64, *const c_void);
 
 pub type ErrorCallback =
-    unsafe extern "C" fn(u32, *const c_void, *const c_void, u32, *const c_void);
+    unsafe extern "C" fn(PxErrorCode::Enum, *const c_void, *const c_void, u32, *const c_void);
 
 extern "C" {
     pub fn physx_create_foundation() -> *mut PxFoundation;
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -354,8 +354,7 @@ pub type ZoneEndCallback = unsafe extern "C" fn(*const c_void, *const i8, bool, 
 pub type ErrorCallback =
     unsafe extern "C" fn(u32, *const c_void, *const c_void, u32, *const c_void);
 
-pub type ErrorUserdataDeleteCallback =
-    unsafe extern "C" fn(*mut c_void);
+pub type ErrorUserdataDeleteCallback = unsafe extern "C" fn(*mut c_void);
 
 extern "C" {
     pub fn physx_create_foundation() -> *mut PxFoundation;
@@ -399,9 +398,7 @@ extern "C" {
         userdata_deleter: ErrorUserdataDeleteCallback,
     ) -> *mut PxErrorCallback;
 
-    pub fn destroy_error_callback(
-        error_callback: *mut PxErrorCallback,
-    );
+    pub fn destroy_error_callback(error_callback: *mut PxErrorCallback);
 
     pub fn get_default_simulation_filter_shader() -> *mut c_void;
 
--- a/physx-sys/src/lib.rs
+++ b/physx-sys/src/lib.rs
@@ -351,6 +351,12 @@ pub type ZoneStartCallback =
 
 pub type ZoneEndCallback = unsafe extern "C" fn(*const c_void, *const i8, bool, u64, *const c_void);
 
+pub type ErrorCallback =
+    unsafe extern "C" fn(u32, *const c_void, *const c_void, u32, *const c_void);
+
+pub type ErrorUserdataDeleteCallback =
+    unsafe extern "C" fn(*mut c_void);
+
 extern "C" {
     pub fn physx_create_foundation() -> *mut PxFoundation;
     pub fn physx_create_foundation_with_alloc(
@@ -360,6 +366,7 @@ extern "C" {
 
     pub fn get_default_allocator() -> *mut PxDefaultAllocator;
     pub fn get_default_error_callback() -> *mut PxDefaultErrorCallback;
+    pub fn create_default_error_callback() -> *mut PxDefaultErrorCallback;
 
     /// Destroy the returned callback object using PxQueryFilterCallback_delete.
     pub fn create_raycast_filter_callback(
@@ -385,6 +392,16 @@ extern "C" {
     ) -> *mut PxProfilerCallback;
 
     pub fn get_alloc_callback_user_data(alloc_callback: *mut PxAllocatorCallback) -> *mut c_void;
+
+    pub fn create_error_callback(
+        error_callback: ErrorCallback,
+        userdata: *mut c_void,
+        userdata_deleter: ErrorUserdataDeleteCallback,
+    ) -> *mut PxErrorCallback;
+
+    pub fn destroy_error_callback(
+        error_callback: *mut PxErrorCallback,
+    );
 
     pub fn get_default_simulation_filter_shader() -> *mut c_void;
 
--- a/physx-sys/src/physx_api.cpp
+++ b/physx-sys/src/physx_api.cpp
@@ -226,6 +226,34 @@ public:
     void *mUserData;
 };
 
+using ErrorCallback = void (*)(PxErrorCode::Enum code, const char* message, const char* file, int line, void* userdata);
+using ErrorUserdataDeleteCallback = void (*)(void* userdata);
+
+class ErrorTrampoline : public PxErrorCallback {
+public:
+    ErrorTrampoline(ErrorCallback errorCb, void* userdata, ErrorUserdataDeleteCallback userdataDeleter)
+        : mErrorCallback(errorCb), mUserdata(userdata), mUserdataDeleteCallback(userdataDeleter) {
+    }
+
+    ~ErrorTrampoline() override {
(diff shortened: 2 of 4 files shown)


physx-sys/src/physx_api.cpp
```diff
@@ -245,13 +273,20 @@ extern "C"
         return &gAllocator;
     }
 
+    // TODO (nises): should this be removed?
```
jnises, line 276: I don't think anything in physx-rs uses it. But it seems to be exported?

physx/src/foundation.rs
```diff
@@ -281,3 +313,42 @@ unsafe impl AllocatorCallback for DefaultAllocator {
...
+    });
+    let message =
+        String::from_utf8_lossy(unsafe { CStr::from_ptr(message.cast::<i8>()) }.to_bytes());
+    let file = String::from_utf8_lossy(unsafe { CStr::from_ptr(file.cast::<i8>()) }.to_bytes());
+    let ec = unsafe { &*user_data.cast::<Box<dyn ErrorCallback>>() };
+    ec.report_error(code, &message, &file, line);
+}
+
+unsafe extern "C" fn error_userdata_drop_helper(user_data: *mut c_void) {
+    let b: Box<Box<dyn ErrorCallback>> =
```
jnises, line 336: some funky stuff here. but it should be correct right?

jnises:
I'm getting warnings that I don't have unsafe in my unsafe functions. But clippy warnings when I do have unsafe in my unsafe functions.
Which way to we want it?

repi:
> I'm getting warnings that I don't have unsafe in my unsafe functions. But clippy warnings when I do have unsafe in my unsafe functions.

for our lints one need to declare `unsafe` blocks inside `unsafe` functions.  maybe you are overriding `RUSTFLAGS` so clippy is not run with the right lints?

tgolsson:
@jnises The clippy settings differ when you run our internal CI or build our internal tools. Physx-RS doesn't require unsafe-in-unsafe because everything in the wrapper is unsafe. When you build with a local path the parent clippy settings are applied to subprojects, which is frustrating. 

But for this project: no unsafe-in-unsafe. Ignore local clippy for that (or temporarily diverge the settings). :-)

repi:
ah I see now, the `unsafe_op_in_unsafe_fn` lint that we are used to using and use in some of our internal repos is not yet in our standard lint set that this repo uses, so currently in this repo one should not have `unsafe` blocks in `unsafe` functions.

though one could easily add a `#![deny(unsafe_op_in_unsafe_fn)]` manually in `lib.rs` here to get the behavior we are used to, unless that is a very large change as is a lot of unsafe in this repo

jnises:
I removed the unnecessary unsafe.

physx/src/foundation.rs
```diff
@@ -84,11 +93,33 @@ pub trait Foundation: Class<physx_sys::PxFoundation> + Sized {
...
+                .cast::<Self>(),
+            )
+        }
+    }
+
+    /// Tries to create a PxFoundation with the provided allocator and error callbacks.
+    /// Returns `None` if `phys_PxCreateFoundation` returns a null pointer.
+    fn with_allocator_error_callback(
+        allocator: Self::Allocator,
+        error_callback: Box<dyn ErrorCallback>,
```
tgolsson, line 112: I'm not super-enthused about this box/double-box thing we get going here. I realize the monomorphization is terrible for code-explosion, but this feels kludgy. Can we do something like convert it into a *PxErrorCallback in the builder already instead?

physx/src/foundation.rs
tgolsson: See my comments above - not sure if we care about dealloc, but this doesn't feel good.

physx/src/foundation.rs
```diff
@@ -281,3 +312,42 @@ unsafe impl AllocatorCallback for DefaultAllocator {
...
+    let code = BitFlags::from_bits(code).unwrap_or_else(|_| {
+        debug_assert!(false, "bad error code {}", code);
+        Default::default()
+    });
+    let message =
+        String::from_utf8_lossy(unsafe { CStr::from_ptr(message.cast::<i8>()) }.to_bytes());
+    let file = String::from_utf8_lossy(unsafe { CStr::from_ptr(file.cast::<i8>()) }.to_bytes());
+    let ec = unsafe { &*user_data.cast::<Box<dyn ErrorCallback>>() };
+    ec.report_error(code, &message, &file, line);
+}
```
tgolsson, line 331: Can we use the regular trampoline pattern instead? Not sure what this code does for us.

physx/src/foundation.rs
```diff
@@ -281,3 +312,42 @@ unsafe impl AllocatorCallback for DefaultAllocator {
...
+        create_error_callback(
+            report_error_helper,
+            // need the extra box to get a normal non-fat pointer
+            Box::into_raw(Box::new(cb)).cast::<c_void>(),
+            error_userdata_drop_helper,
+        )
+    }
+}
+
+pub trait ErrorCallback {
```
tgolsson, line 347: A small side note here: let's bring all code related to the ErrorCallback into a submodule; avoids cluttering the foundation.rs file with helpers. It's fine to re-export the relevant types from here; like I do with profiler.

tgolsson (changes requested):
Hey, thanks for the PR! Had a look through it, and I'm not super-enthused about the approach here since it breaks with our pattern. I can see why you've gone this route since generalization over all types is quite frustrating. 

I'd much rather do it in another way by moving the conversion-to-px-type (or a type-erased wrapper) further out to the builder API. If the builder took <EC: ErrorCallback> and immediately did the `ec.into_px()` we avoid the sizeness problems. 

Overall looks good, just some ergonomics, maintainability concerns!

physx-sys/src/lib.rs
```diff
@@ -360,6 +365,7 @@ extern "C" {
 
     pub fn get_default_allocator() -> *mut PxDefaultAllocator;
     pub fn get_default_error_callback() -> *mut PxDefaultErrorCallback;
+    pub fn create_default_error_callback() -> *mut PxDefaultErrorCallback;
```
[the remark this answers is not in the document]
tgolsson: (OK, read later on. See note re: dealloc).

physx-sys/src/lib.rs
```diff
@@ -386,6 +392,14 @@ extern "C" {
 
     pub fn get_alloc_callback_user_data(alloc_callback: *mut PxAllocatorCallback) -> *mut c_void;
 
+    pub fn create_error_callback(
+        error_callback: ErrorCallback,
+        userdata: *mut c_void,
+        userdata_deleter: ErrorUserdataDeleteCallback,
```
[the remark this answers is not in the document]
jnises: Yeah. This would would be much simpler if we just leak everything.
It feels kinda bad to do that in a library like this though. Even if this allocation shouldn't matter much here. What if someone else uses the library in some other way?

physx/src/foundation.rs
jnises: I remember trying to do this using generics, but it got too complicated.
I can try again.

The double box though. Won't that still be needed at some level? Since our ffi is using void pointers for the user data?
Gonna have to fit that fat pointer into a usize somehow. Or is there some way to do vtables c++ style?
Perhaps I can make user data fit a fat pointer? But that wouldn't fit with our existing pattern either..

physx/src/foundation.rs
jnises: what would the regular trampoline pattern look like?

tgolsson: So if you look at e.g. the profiler trampoline, they don't use any helper functions like this. They replicate the C++ interface in a Rust trait, with an additional userdata at the end. Dealing with type conversions is left to the user of the library, because they might want to do different things, be on different platforms, etc.

physx/src/foundation.rs
tgolsson: With how the other trampolines are implemented (without box/dyn Trait), they don't have to pass a fat pointer - they end up passing a pointer to the concrete type. So we end up with a userdata +fn pointer trampoline. Pretty much a vtable.

physx-sys/src/physx_api.cpp
tgolsson: One can use physx-sys directly, which makes this required.

physx-sys/src/lib.rs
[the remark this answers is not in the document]
tgolsson: If we want to fix this, we should fix all of them in one go.

physx/src/foundation.rs
jnises: I figured it would be nice to have physx-rs provide safe apis and leave the pointer stuff to physx-sys as much as possible?

tgolsson: Hmm. Yes. But I'm not sure if that should come at the cost of a double indirection. Wondering whether we could skip one layer of dynamic dispatch somehow...

tgolsson: Maybe we could do this using a monomorphized shim instead, which would get rid of any dynamic calls here.

```
trait PhysxLoggingCallback {
    fn on_message<'a>(&self, msg: &'a str);
}


struct LoggingVTable {
    on_message: fn(state: *const c_void, msg: *const i8);
}

fn into_vtable<L: PhysxLoggingCallback>(callback: L) -> LoggingVtable {
    fn on_message_shim<L: PhysxLoggingCallback>(this: *const c_void, msg: *const i8) {
        let this = unsafe { &*this.cast::<L>() };
        let msg = OsStr::from_raw(msg).unwrap().as_str();
        this.on_message(msg);
    }
    
    LoggingVTable {
        on_message: on_message_shim<L>
    }
}
```

tgolsson: (untested code, just off the top of my head for how this pattern generally looks)

jnises: nice. yeah, that should avoid one indirection.
I'll try and do something like that.

tgolsson: @jnises Sure! LMK if you want me to help out more actively :-)

physx-sys/src/lib.rs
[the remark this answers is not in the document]
jnises: so the problem here is the `userdata_deleter`? or is it the existence of the `create_error_callback` function?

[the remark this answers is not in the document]
tgolsson: The deletion part. I think let's focus on the logging callback; and we can ensure all resources can be deleted in the next step.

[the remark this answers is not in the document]
jnises: I'll have one more go at cleaning up the deletion part.
If that doesn't work I'll make a leaky version.
I just worry that it'll be hard to retrofit the deletion without introducing breaking changes to the api..

physx/src/foundation.rs
```diff
@@ -92,6 +94,25 @@ pub trait Foundation: Class<physx_sys::PxFoundation> + Sized {
         }
     }
 
+    /// Tries to create a PxFoundation with the provided allocator and error callbacks.
+    /// `error_callback` must live as long as the returned `Foundation`
```
jnises, line 98: Is it ok to just document these things like this?
Or is there backwards compatible way to enforce it?

repi: yes any `unsafe` function should ideally have a doc line that describes the function and then a `# Safety` section where the unsafe invariants are described like this. There is even a lint for that that we ideally should use in this repo (and do use in ark)

repi: example: https://rust-lang.github.io/api-guidelines/documentation.html#c-failure

jnises: Thanks.
I added safety sections

physx-sys/src/lib.rs
```diff
@@ -351,6 +351,11 @@ pub type ZoneStartCallback =
 
 pub type ZoneEndCallback = unsafe extern "C" fn(*const c_void, *const i8, bool, u64, *const c_void);
 
+pub type ErrorCallback =
+    unsafe extern "C" fn(u32, *const c_void, *const c_void, u32, *const c_void);
```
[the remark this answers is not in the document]
jnises: I see.

It seems `PxErrorCode::Enum`  has type `i32` in the generated code.
Why is that?
It seems most other enums are u32.

repi (approved):
accepting just not to block this on my previous PR comments that have been resolved

physx-sys/src/lib.rs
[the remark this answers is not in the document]
jnises: ok

tgolsson:
Superseded by #168 after discussion on Slack.

PR #168, opened 2022-10-18
Trait + early conversion for error callback

tgolsson:
A slight modification to make it work a bit more like your original proposal; while still keeping the trait pattern we have. Also avoids the sized-ness/boxing issue we've seen before.

Not sure why the char types were *const c_void; we try to keep them as signed chars in the interface as it's most consistent.

### DIFF
--- a/physx/src/foundation.rs
+++ b/physx/src/foundation.rs
@@ -16,10 +16,11 @@ use physx_sys::{
     PxFoundation_getAllocatorCallback_mut, PxFoundation_getErrorCallback_mut,
     PxFoundation_getErrorLevel, PxFoundation_getReportAllocationNames, PxFoundation_release_mut,
     PxFoundation_setErrorLevel_mut, PxFoundation_setReportAllocationNames_mut,
+    create_error_callback, destroy_error_callback, create_default_error_callback,
 };
 use std::{
     alloc::{alloc, dealloc, Layout},
-    ffi::c_void,
+    ffi::{c_void, CStr},
     marker::PhantomData,
     mem::{align_of, size_of},
     sync::atomic::{AtomicUsize, Ordering::SeqCst},
@@ -39,6 +40,8 @@ pub enum ErrorCode {
     PerfWarning = 128u32,
 }
 
+pub type ErrorCodeFlags = BitFlags<ErrorCode>;
+
 /// A new type wrapper for PxFoundation.  Parametrized by it's Allocator type.
 #[repr(transparent)]
 pub struct PxFoundation<Allocator: AllocatorCallback> {
@@ -52,7 +55,11 @@ impl<Allocator: AllocatorCallback> Drop for PxFoundation<Allocator> {
             if let Some(allocator) = self.get_allocator_callback() {
                 drop(Box::from_raw(allocator));
             };
-            PxFoundation_release_mut(self.as_mut_ptr())
+            let error_callback: *mut PxErrorCallback = self.get_error_callback().map(|x| x as *mut PxErrorCallback).unwrap_or(std::ptr::null_mut());
+            PxFoundation_release_mut(self.as_mut_ptr());
+            if !error_callback.is_null() {
+                destroy_error_callback(error_callback);
+            }
         }
     }
 }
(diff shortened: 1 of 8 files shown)


physx/src/physics.rs
```diff
@@ -736,11 +738,11 @@ impl<Allocator: AllocatorCallback> PhysicsFoundationBuilder<Allocator> {
...
     /// # Safety
     /// `error_callback` must live as long as the object created by `build`
-    pub unsafe fn with_error_callback(
+    pub unsafe fn with_error_callback<EC: ErrorCallback>(
         &mut self,
-        error_callback: *mut PxErrorCallback,
+        error_callback: EC,
     ) -> &mut Self {
-        self.error_callback = Some(error_callback);
+        self.error_callback = Some(error_callback.into_px());
```
jnises, line 745: so we leak the object here, and the boxed self below.
is that ok?

physx/src/physics/error_callback.rs
```diff
@@ -0,0 +1,42 @@
...
+        unsafe extern "C" fn on_message_shim<L: ErrorCallback>(
+            code: physx_sys::PxErrorCode::Enum,
+            message: *const i8,
+            file: *const i8,
+            line: u32,
+            this: *const c_void,
+        ) {
+            let this = &*this.cast::<L>();
+            let msg = CStr::from_ptr(message.cast());
+            let msg = msg.to_str().unwrap(); // SAFETY: We're dealing with basic char* here, so should always convert cleanly to str.
```
jnises, line 23: not if it doesn't contain valid utf8?

physx/src/physics/error_callback.rs
```diff
@@ -0,0 +1,42 @@
...
+
+            let file = CStr::from_ptr(file.cast());
+            let file = file.to_str().unwrap(); // SAFETY: We're dealing with basic char* here, so should always convert cleanly to str.
+
+            let code = if code == -1 {
```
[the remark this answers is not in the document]
tgolsson: Nope! i32(-1) is all bits set; and only some of those bits map to variants in the flagset!

physx/src/physics/error_callback.rs
tgolsson: I'm making two assumptions to say that isn't an issue: All data passed into PhysX from this library is already UTF-8 so it cannot be broken, and all formatted strings in PhysX should contain only ASCII. If we can find a nice way to handle this without silently ignoring errors I'm fine with being stricter/more careful.

physx/src/physics.rs
tgolsson: Yes; for now. :-) As noted on Slack we can solve this at a later point without any API changes.

physx/src/physics/error_callback.rs
jnises: Yeah, it probably won't ever be an issue.

I guess you could imagine a situation where someone for some reason decide to compile physx in a directory containing non-ascii characters, on linux, with the system encoding set to something other than utf8. Then `__FILE__` could contain non-ascii chars, which would end up in `file` :)

jnises: but perhaps at least change the SAFETY comment? "basic char*" isn't really a guarantee for anything

tgolsson: Switched to `to_string_lossy` unconditionally, PhysX doesn't log enough for it to matter.

physx/src/physics.rs
```diff
@@ -711,9 +735,21 @@ impl<Allocator: AllocatorCallback> PhysicsFoundationBuilder<Allocator> {
         self
     }
 
+    /// Set error callback
+    /// # Safety
+    /// `error_callback` must live as long as the object created by `build`
```
jnises, line 740: this safety note is no longer needed right?

physx/src/foundation.rs
```diff
@@ -84,14 +84,31 @@ pub trait Foundation: Class<physx_sys::PxFoundation> + Sized {
     /// Returns `None` if `phys_PxCreateFoundation` returns a null pointer.
     fn new(allocator: Self::Allocator) -> Option<Owner<Self>> {
         unsafe {
-            Owner::from_raw(phys_PxCreateFoundation(
-                crate::physics::PX_PHYSICS_VERSION,
-                allocator.into_px(),
+            Self::with_allocator_error_callback(
+                allocator,
                 get_default_error_callback() as *mut PxErrorCallback,
```
jnises, line 89: If we want to handle deallocattion of the callback later. How will we tell the difference between this singleton, and a pointer created by `create_error_callback`?

physx/src/physics.rs
tgolsson: This is still a public API so I think it makes sense to annotate an unsafe function even if our preferred method is fully safe.

jnises: but this method isn't unsafe anymore?

tgolsson: Oh! Sorry! Yes. 🤦

physx/src/physics.rs
```diff
@@ -711,9 +735,18 @@ impl<Allocator: AllocatorCallback> PhysicsFoundationBuilder<Allocator> {
         self
     }
 
+    pub fn with_error_callback<EC: ErrorCallback>(&mut self, error_callback: EC) -> &mut Self {
+        self.error_callback = unsafe { Some(error_callback.into_px()) };
```
jnises, line 739: Add todo about deallocating here so we don't forget?

physx/src/physics/error_callback.rs
```diff
@@ -0,0 +1,43 @@
...
+                BitFlags::from_bits(code as u32)
+                    .unwrap_or_else(|_| ErrorCode::InvalidParameter.into())
+            };
+
+            this.report_error(code, &msg, &file, line);
+        }
+
+        create_error_callback(
+            on_message_shim::<Self>,
+            Box::into_raw(Box::new(self)) as *mut c_void,
```
jnises, line 40: Add todo about deallocating here so we don't forget?

tgolsson: https://github.com/EmbarkStudios/physx-rs/issues/169 instead.

physx/src/physics.rs
tgolsson: Created an issue.

jnises (approved):

PR #176, opened 2022-11-22
Update

Jake-Shadle:
Fixes some bitwise instead of logical warnings, and disables warnings as errors for the C++ code. This is not useful for end users since they don't have direct control of the C++ code, and just results in build failures for new compiler versions.


tgolsson (changes requested):
Changelog needed but otherwise LGTM! Can't approve due to overeager bots. :-)

PR #182, opened 2023-02-16
Fix some clippy lints from 1.66 and switch to newer XCode

tgolsson:


.github/workflows/ci.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04, macOS-latest]
     env:
-      DEVELOPER_DIR: /Applications/Xcode_11.7.app
+      DEVELOPER_DIR: /Applications/Xcode_14.2.app
     runs-on: ${{ matrix.os }}
```
Jake-Shadle, line 48: I don't even know why we're setting this, maybe it's better to just remove.

Jake-Shadle (approved):

Issue #175, opened 2022-11-08
Port to PhysX 5


Jake-Shadle:
https://github.com/NVIDIA-Omniverse/PhysX

Should mostly be mechanical changes, though we'll also have to port our fixes for cross-compilation since those are (I assume) still broken in the new one since they never responded to our PR.

Jake-Shadle:
So unfortunately physx 5 changed some important things, like all of the core math types being converted (unnecessarily) to templates, and some other changes making pxbind no longer function (possibly due to using 15 instead). So I've started porting pxbind to rust, and instead of using the clunky libclang, just dumping the ast and parsing it, which still gives us all the info we need, but in a form that is easier to use with rust.

PR #183, opened 2023-02-17
Upgrade to PhysX 5

Jake-Shadle:
## pxbind

This ports pxbind from C++ using libclang -> rust using clang's AST JSON output. The port includes adding snapshot testing of the output of pxbind + structgen both to validate the port as well as (hopefully) making any changes easier in the future. This initial port is _mostly_ a 1 to 1 conversion with 3 notable exceptions.

- [x] `enums` **improved**
- [x] `flags` **improved**
- [x] `core types` **improved**
- [x] `methods`
- [x] `structs/classes`
- [x] ~templates - Unfortunately more templates have crept into the public API of PhysX, to a degree where it feels like we need to support at least some of them.~

### `enums`

In the current pxbind, enums are outputted like this.

```rust
pub mod PxShapeFlag{
pub type Enum = u32;
pub const eSIMULATION_SHAPE: Enum = 1u64 as u32;
pub const eSCENE_QUERY_SHAPE: Enum = 2u64 as u32;
pub const eTRIGGER_SHAPE: Enum = 4u64 as u32;
pub const eVISUALIZATION: Enum = 8u64 as u32;
}
```

This makes them cumbersome to use from Rust. In the new pxbind they are outputted like this.

```rust
/// Flags which affect the behavior of PxShapes.
#[derive(Copy, Clone)]
#[repr(u32)]
pub enum PxShapeFlag {
    /// The shape will partake in collision in the physical simulation.
    ///
    /// It is illegal to raise the eSIMULATION_SHAPE and eTRIGGER_SHAPE flags.
    /// In the event that one of these flags is already raised the sdk will reject any
    /// attempt to raise the other.  To raise the eSIMULATION_SHAPE first ensure that
    /// eTRIGGER_SHAPE is already lowered.
    ///
    /// This flag has no effect if simulation is disabled for the corresponding actor (see [`PxActorFlag::eDISABLE_SIMULATION`]).
    eSIMULATION_SHAPE = 1,
    /// The shape will partake in scene queries (ray casts, overlap tests, sweeps, ...).
    eSCENE_QUERY_SHAPE = 2,
    /// The shape is a trigger which can send reports whenever other shapes enter/leave its volume.
    ///
    /// Triangle meshes and heightfields can not be triggers. Shape creation will fail in these cases.
    ///
    /// Shapes marked as triggers do not collide with other objects. If an object should act both
    /// as a trigger shape and a collision shape then create a rigid body with two shapes, one being a
    /// trigger shape and the other a collision shape. 	It is illegal to raise the eTRIGGER_SHAPE and
    /// eSIMULATION_SHAPE flags on a single PxShape instance.  In the event that one of these flags is already
    /// raised the sdk will reject any attempt to raise the other.  To raise the eTRIGGER_SHAPE flag first
    /// ensure that eSIMULATION_SHAPE flag is already lowered.
    ///
    /// Trigger shapes will no longer send notification events for interactions with other trigger shapes.
    ///
    /// Shapes marked as triggers are allowed to participate in scene queries, provided the eSCENE_QUERY_SHAPE flag is set.
    ///
    /// This flag has no effect if simulation is disabled for the corresponding actor (see [`PxActorFlag::eDISABLE_SIMULATION`]).
    eTRIGGER_SHAPE = 4,
    /// Enable debug renderer for this shape
    eVISUALIZATION = 8,
}
```

This allows regular usage from Rust, and the Rust enum can be used on the FFI boundary since it is guaranteed to be the same size as the underlying integer type in C++.

### `flags`

Many enums in PhysX also have a bitflags version using `PxFlags<_enum_type_, _integer_type_>`.

`typedef PxFlags<PxShapeFlag::Enum,PxU8> PxShapeFlags;`

In the current pxbind, they are outputted like this.

```rust
#[derive(Clone, Copy)]
#[repr(C)]
pub struct PxShapeFlags{
    pub mBits: u8,
}
```

In the new pxbind they are outputted like this.

```rust
bitflags::bitflags! {
    /// Flags for [`PxShapeFlag`]
    #[derive(Copy, Clone, Default)]
    #[repr(transparent)]
    pub struct PxShapeFlags: u8 {
        const eSIMULATION_SHAPE = 1 << 0;
        const eSCENE_QUERY_SHAPE = 1 << 1;
        const eTRIGGER_SHAPE = 1 << 2;
        const eVISUALIZATION = 1 << 3;
    }
}
```

Like the regular enums, this allows easy and typesafe usage from Rust that can be passed transparently through FFI. Since the enum that is paired with the bitflags are always emitted before the flags we elide the comments on the individual constants.

### Core Types

~PxVec3, PxVec4, PxVec3Padded, PxQuat, PxMat33, PxMat44, and PxTransform now use the equivalent `glam` types on FFI since they all already match the layout of the PhysX pod types. This means we also skip generating FFI bindings for the various math functions related to those types as that can just be done in pure Rust without needing to go back and forth. This was also done because PhysX now makes all of these types into templates, which makes everything more annoying, especially since the f64 versions are...never actually used. But future proof!~

I've changed the core types back to non-templated versions, there is no reason to keep that baggage. https://github.com/EmbarkStudios/PhysX-5/commit/87ee3fa1f42f36f4efacf920af12c7f94e708fe2

## physx-sys

Since physx-sys just exposes the raw bindings generated by pxbind/structgen the breaking changes are mostly the same as those in the underlying sdk, eg things like constructors in core types like PxTransform now take defaults, removing 1 or more constructor bindings which can cause direct use to be incorrect due to how constructors are named with a number suffix to disambiguate the FFI functions. There were also types that were completely removed such as PxArticulation and so forth.

## physx

In addition to the aforementioned changes in physx-sys, probably the biggest change in physx is that we no longer have to manually recreate bit flags and do the conversion to/from the FFI type since _most_ bitfields (there are a few exceptions) are now automatically generated by pxbind.

Resolves: #175


Jake-Shadle:
Note I'm not expecting a thorough review of this PR due to its size, but I'll summarize some of the important bits.

physx-sys/pxbind/tests/snapshots/functions__plain_functions-2.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: physx-sys/pxbind/tests/functions.rs
+expression: mo.rust
+---
+extern "C" {
+}
+
```
tgolsson, line 7: I've not found the test code here yet (I think) but is this right?

tgolsson (commented):
Got a bit left to go; leaving this here for today. Very nice work! I've mostly skimmed but saw some tests that looked like they generate nothing  for both Rust/CPP. 

Also some scripts that I think can do with some docs 🙇 

Got the generator left mostly, want to do that on a fresh brain tomorrow.

physx-sys/build.rs
```diff
@@ -335,7 +339,7 @@ fn add_common(ctx: &mut Context) {
 
     let flags = if builder.get_compiler().is_like_clang() {
         vec![
-            "-std=c++14",
+            "-std=c++11",
```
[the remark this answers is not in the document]
Jake-Shadle: Yes.

physx-sys/migration-4-5.md
```diff
@@ -0,0 +1,272 @@
...
+- `PxScene` queries have _mostly_ been moved to [`PxSceneQueryExt`](https://github.com/EmbarkStudios/PhysX-5/blob/4d33492253ab64c2246d80257ac4ce0f37b40b22/physx/include/extensions/PxSceneQueryExt.h#L60)
+- `PxRigidBody_setAngularVelocity_mut` and `PxRigidBody_setLinearVelocity_mut` have been moved to [`PxRigidDynamic`](https://github.com/EmbarkStudios/PhysX-5/blob/main/physx/CHANGELOG.md#changed-3)
+- Many class constructors have been consolidated, so instead of having, for example, a default constructor and a constructor that takes two `f32` parameters, there will now be single constructor that takes two `f32` parameters, both having default values that used to be set in the default constructor. If you were calling a `_new` function (a wrapper for a constructor) that now fails since the signature is different, you can look at the code to determine what arguments to pass to retain the default constructor behavior.
+
+```diff
+-pub fn PxCapsuleGeometry_new() -> PxCapsuleGeometry;
++pub fn PxCapsuleGeometry_new(radius_: f32, halfHeight_: f32) -> PxCapsuleGeometry;
+```
+
+[PR#183]: https://github.com/EmbarkStudios/physx-rs/pull/183
```
tgolsson, line 272: Awesome docs 😍

physx/src/articulation_link.rs
```diff
@@ -115,53 +88,62 @@ pub trait ArticulationLink: Class<physx_sys::PxArticulationLink> + RigidBody + U
...
+    // Get inbound joint for this link
+    // #[inline]
+    // fn get_inbound_joint(&self) -> Option<&JointMap> {
+    //     unsafe {
+    //         (&PxArticulationLink_getInboundJoint(self.as_ptr())
+    //             as *const *mut physx_sys::PxArticulationJointReducedCoordinate
+    //             as *const JointMap)
+    //             .as_ref()
+    //     }
+    // }
```
tgolsson, line 125: Gone or just broken?

physx-sys/update-physx.sh
```diff
@@ -1,18 +1,22 @@
...
+fi
+
+pushd build
+cmake -G "Unix Makefiles" ..
+make -j32
+popd
+
+export PHYSX_ROOT_DIR
+
+./run.sh "$1"
```
tgolsson, line 22: What args does this take? Where should the main script be ran from?

tgolsson (commented):
The new codegen tooling looks fantastic, such an improvement. Only some minor questions left from my side.

I'm noticing we still miss some generated files as well; I can only see the linux_x86_64 ones. I assume the other ones will show up, just wanted to point it out so we don't miss it. 

Awesome work!

Jake-Shadle:
> I'm noticing we still miss some generated files as well; I can only see the linux_x86_64 ones. I assume the other ones will show up, just wanted to point it out so we don't miss it.

I'm generating all the other platforms at the moment.

physx/src/articulation_link.rs
Jake-Shadle: The function is still there, but since I removed JointMap since all the base classes were removed I just commented this out for now, can add it back in a later version.

Jake-Shadle:
Ok, all platforms that we care about are about are compiling and working again, I'll merge this tomorrow. It's time to go and drink heavily.

tgolsson:
@Jake-Shadle Sounds good. I'll leave you to override the merge so you don't get raced by the bot.

Issue #187, opened 2023-03-01
Port PxCooking


Jake-Shadle:
5.1 deprecated PxCooking in favor of just using the lower level functions, I undeprecated it in https://github.com/EmbarkStudios/PhysX-5/commit/5cdbcf6c0ff9c9ff60397c909227243ea0fd8177 temporarily until we do that change.

PR #191, opened 2023-03-03
Replace PxCooking with non-deprecated functions

Jake-Shadle:
- Fix DeriveClassForNewType macro
- Replace PxCooking with non-deprecated functions
- Cleanup regen

Resolves: #187


tgolsson (commented):
Changelog, otherwise LGTM!

PR #193, opened 2023-03-07
Fix unnecessary recompiles

Jake-Shadle:
The `PxPhysicsVersion.h` file moved (both because we changed the submodule checkout location, but also because of an upstream change), and instead of giving a warning that the file didn't exist, cargo just...rebuilt the physx-sys build code every time.


tgolsson (approved):
Thank you!

PR #201, opened 2023-07-06
Add support for post filter

NiklasNummelin:
Implements a C++ trampoline and Rust interface for setting both pre and post filters.


tgolsson:
Changelogs please, and green CI :-)

NiklasNummelin:
Done! :) @tgolsson

tgolsson (approved):
LGTM. Since renaming would be a breaking change we can do that at a later point.

Issue #205, opened 2023-07-14
ActorMap is missing read-only API


rlidwka:
### Description of the problem

I'm trying to register `on_wake_sleep` callback, which has the following signature:

```rust
pub trait WakeSleepCallback<L: ArticulationLink, S: RigidStatic, D: RigidDynamic>: Sized {
    fn on_wake_sleep(&mut self, actors: &[&ActorMap<L, S, D>], is_waking: bool);
}
```

ActorMap as of now has the following implementation ([link](https://github.com/EmbarkStudios/physx-rs/blob/1c6fb68824de9f9b61ac3fb2afa006e7c572ca0e/physx/src/actor.rs#L143-L199)):

```rust
impl<L, S, D> ActorMap<L, S, D>
where
    L: ArticulationLink,
    S: RigidStatic,
    D: RigidDynamic,
{
    pub fn cast_map<'a, Ret, ALFn, RSFn, RDFn>(
        &'a mut self,
        mut articulation_link_fn: ALFn,
        mut rigid_static_fn: RSFn,
        mut rigid_dynamic_fn: RDFn,
    ) -> Ret {/*...*/}

    pub fn as_rigid_dynamic(&mut self) -> Option<&mut D> {/*...*/}
    pub fn as_rigid_static(&mut self) -> Option<&mut S> {/*...*/}
    pub fn as_articulation_link(&mut self) -> Option<&mut L> {/*...*/}
}
```

Please notice that every single method in `ActorMap` requires `&mut self`, but `WakeSleepCallback` only gives access to `&self`, thus making wake_sleep callback unusable as is.

*Any physx user can of course cast pointers by herself similar to how ActorMap does it internally, but that's not ideal.*

### Proposed solution

1. rename existing `cast_map` and `as_*` to `cast_map_mut` and `as_*_mut` respectively.
2. add `cast_map` and `as_*` functions which take `&self` and return `Option<&self>`

For example:

```rust
// instead of this
pub fn as_rigid_dynamic(&mut self) -> Option<&mut D> {/*...*/}

// suggesting to do this
pub fn as_rigid_dynamic(&self) -> Option<&D> {/*...*/}
pub fn as_rigid_dynamic_mut(&mut self) -> Option<&mut D> {/*...*/}
```

rlidwka:
> Seems sensible for me as described here -- feel free to open a PR with the changes.

done

[...]

PR #206, opened 2023-07-17
Add immutable API for ActorMap

rlidwka:
see discussion in https://github.com/EmbarkStudios/physx-rs/issues/205


tgolsson (approved):
Sorry for taking a while, lost this during vacations! Thanks for the contribution!

Issue #210, opened 2023-08-29
Misaligned pointer dereference


fgadaleta:
**Describe the bug**

```
2023-08-29T08:14:34.620649Z  INFO bevy_winit::system: Creating new window "Bevy App" (0v0)
2023-08-29T08:14:34.621476Z  INFO winit::platform_impl::platform::x11::window: Guessed window scale factor: 1    
2023-08-29T08:14:35.225130Z  INFO bevy_render::renderer: AdapterInfo { name: "NVIDIA GeForce RTX 3060 Ti", vendor: 4318, device: 9353, device_type: DiscreteGpu, driver: "NVIDIA", driver_info: "525.125.06", backend: Vulkan }
thread 'main' panicked at /home/frag/.cargo/git/checkouts/bevy_mod_physx-ec641f3181f0becd/467a452/src/physx/src/traits/user_data.rs:26:41:
misaligned pointer dereference: address must be a multiple of 0x8 but is 0x7ffff4a09606
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4e78abb437a0478d1f42115198ee45888e5330fd/library/std/src/panicking.rs:619:5
   1: core::panicking::panic_nounwind_fmt
             at /rustc/4e78abb437a0478d1f42115198ee45888e5330fd/library/core/src/panicking.rs:106:14
   2: core::panicking::panic_misaligned_pointer_dereference
             at /rustc/4e78abb437a0478d1f42115198ee45888e5330fd/library/core/src/panicking.rs:193:5
   3: physx::traits::user_data::UserData::init_user_data
             at /home/frag/.cargo/git/checkouts/bevy_mod_physx-ec641f3181f0becd/467a452/src/physx/src/traits/user_data.rs:26:41
   4: physx::material::Material::from_raw
             at /home/frag/.cargo/git/checkouts/bevy_mod_physx-ec641f3181f0becd/467a452/src/physx/src/material.rs:122:25
   5: physx::physics::Physics::create_material
             at /home/frag/.cargo/git/checkouts/bevy_mod_physx-ec641f3181f0becd/467a452/src/physx/src/physics.rs:418:13
   6: <bevy_physx::PhysicsCore as bevy_app::plugin::Plugin>::finish
             at /home/frag/.cargo/git/checkouts/bevy_mod_physx-ec641f3181f0becd/467a452/src/lib.rs:334:22
   7: bevy_app::app::App::finish
[...]

PR #211, opened 2023-08-30
Fix misaligned pointer dereference when storing UserData

rlidwka:
see discussion in https://github.com/EmbarkStudios/physx-rs/issues/210

I've also added a test, which verifies userdata storing/retrieving with various types, and may also reproduce a problem with older code. You can run it with the following command:

```sh
cargo test --package physx --lib -- traits::user_data::tests --nocapture
```

line 28 is where the bug is at, lines 24-25 are changed only for consistency reasons

### DIFF
--- a/physx-sys/pxbind/src/consumer.rs
+++ b/physx-sys/pxbind/src/consumer.rs
@@ -299,7 +299,9 @@ impl<'ast> AstConsumer<'ast> {
                     // If a record decl doesn't have any inner nodes, it's just
                     // a foreward declaration and we can skip it
                     if inn.inner.is_empty() || rec.definition_data.is_none() {
-                        let Some(name) = rec.name.as_deref() else { continue; };
+                        let Some(name) = rec.name.as_deref() else {
+                            continue;
+                        };
 
                         if !self.classes.contains_key(name) {
                             self.recs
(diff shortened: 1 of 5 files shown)


tgolsson:
Awesome work! Quick `cargo fmt` and we should be good to go!

rlidwka:
> Quick `cargo fmt` and we should be good to go!

Okay, I've added formatting changes as the second commit.

*Note: `cargo fmt` errors happened because rust people finally figured out "the only right way" to write let..else statements in a recent rust release, and those changes have nothing to do with the changes in the first commit.*

tgolsson:
:+1: Yeah; figures. Thanks for fixing anyways. I'll see if I can cut a release later this week, a bit swamped right now.

tgolsson (approved):

rlidwka:
I've just noticed that other two methods in the same file use **double**-unsafe blocks, i.e.:

```rust
unsafe fn fn get_user_data_mut(...) {
    unsafe {
        ...
    }
}
```

Example:

https://github.com/EmbarkStudios/physx-rs/blob/bbfa160da9e15e80e62440a920503537ea8569ad/physx/src/traits/user_data.rs#L51-L62

What's the reason for that, and should I have wrapped `if..else` in `unsafe` block in a similar way?

tgolsson:
I'm going to guess it comes from this lint: https://doc.rust-lang.org/beta/nightly-rustc/rustc_lint/builtin/static.UNSAFE_OP_IN_UNSAFE_FN.html, though it doesn't seem like we apply it consistently...

PR #212, opened 2023-09-18
Use -fno-strict-aliasing when building physx

forny:
A lot of undefined behavior with all the reinterpret_cast usage in physx.

The following was broken compiling without `-fno-strict-aliasing` with clang/llvm 16:
https://github.com/NVIDIA-Omniverse/PhysX/blob/main/physx/source/geomutils/src/intersection/GuIntersectionRayBox.cpp#L199


tgolsson (approved):
LGTM, thanks!

PR #215, opened 2023-09-27
Do not manually free materials for shapes

### LINKED ISSUE #180: Segfault (double free) on drop(PxShape)
```rs
let mut material = physics.create_material(0.5, 0.5, 0.6, ()).unwrap();
let geometry = PxBoxGeometry::new(1., 1., 1.);
let flags = ShapeFlag::SceneQueryShape | ShapeFlag::SimulationShape | ShapeFlag::Visualization;
let mut shape = physics.create_shape(&geometry, &mut [&mut material], false, flags, ()).unwrap();
drop(material);
drop(shape);
```

This code releases `material` twice. Material itself is an owned pointer. And then shape releases all its materials (see line 100 below):

https://github.com/EmbarkStudios/physx-rs/blob/f0495378d0215179d73976ec3684ecb291f42ec8/physx/src/shape.rs#L96-L106

Either `create_shape` should consume material (i.e. `&mut [Owner<Material>]` instead of `&mut [&mut Material]`) or PxShape shouldn't release its materials on drop. I believe it's latter, but not completely sure.

tgolsson:
Fixes #180

@rlidwka  Correct analysis indeed, Physx internally decrements the ref count so no need for our wrappers to do so -- that leads to decrementing the ref-count twice. Do you have a bigger context where you'd be able to try this and ensure it now works?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,3 @@
 [workspace]
+resolver = "2"
 members = ["physx", "physx-sys", "physx-sys/pxbind"]
--- a/physx/src/material.rs
+++ b/physx/src/material.rs
@@ -140,7 +140,7 @@ pub trait Material: Class<physx_sys::PxMaterial> + UserData {
     }
 
     /// Set the restitution.
-    /// - Restitution must be in [0.0 ..= 1.0], values outside tyhe range are clamped.
+    /// - Restitution must be in [0.0 ..= 1.0], values outside the range are clamped.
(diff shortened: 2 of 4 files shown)


rlidwka:
> Do you have a bigger context where you'd be able to try this and ensure it now works?

I've tried it, and I'm pretty sure this solution works.

tgolsson:
Thanks! Will try to get a release done next week.

bpaberg (approved):
Looks good to me but are the trait implementations needed in this test?

PR #229, opened 2024-02-06
Add Peter and Peter as CODEOWNERS

tgolsson:
Since I don't actually do much with physx/ics we've talked about having @blunzn and @bpaberg take over maintainership of this repository in the long term.


bpaberg (approved):

blunzn (approved):
I feel honoured! :)

